PR #3, opened 2019-01-15
Refactored production of errors in response.

stevieing:
I don't know if it is over engineered. Might be good to add some comments.


harrietc52:
👍very nice

PR #28, opened 2019-02-22
Seq 938 json api deserializer

stevieing:
Merged deserializer, response and request into all of the views.
Refactored some of the methods.
full Lint.
There are still two skipped tests in Modal.spec.js that fail.


harrietc52:
Yep all looks great! Scanned over only in browser, didnt have time to play locally, but like the Api mdule and the successful method. Happy to merge 👍

PR #31, opened 2019-02-26
Creation of new Sequencing Run

harrietc52:


src/views/Runs.vue
```diff
@@ -0,0 +1,96 @@
...
+    <div class="runs">
+      <b-button id="newRun" class="float-right" @click="createNewRun">Create New Run</b-button>
+
+      <b-table
+         show-empty
+         :items="getRuns"
+         :fields="fields"
+      >
+        <template slot="selected" slot-scope="row">
+          <b-checkbox @change="toggleSelectedRow(row.item)"></b-checkbox>
```
stevieing: You can use v-model which negates the need for toggleSelectedRow method. This is also used in all of the other views.

harrietc52: I think I'll remove it from here all together as it isnt being used at the moment

PR #51, opened 2019-03-08
Modified scan barcodes so that it can be used for sequencescape and t…

stevieing:
…raction.

### DIFF
--- a/src/views/ScanBarcodes.vue
+++ b/src/views/ScanBarcodes.vue
@@ -3,7 +3,9 @@
     <div class="form-group">
       <label for="barcodes">barcodes:</label>
       <textarea type="text" v-model="barcodes" class="form-control" rows="10" cols="10" name="barcodes" id="barcodes" />
-      <b-button id="findTubes" variant="success" @click="findTubes" >Go!</b-button>
+      <b-button id="findSequencescapeTubes" variant="success" @click="findSequencescapeTubes" >find Sequencescape Tubes</b-button>
+      <b-button id="findTractionTubes" variant="success" @click="findTractionTubes" >find Traction Tubes</b-button>
+
     </div>
   </div>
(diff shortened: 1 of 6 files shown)


src/views/ScanBarcodes.vue
```diff
@@ -28,31 +30,51 @@ export default {
...
+          this.message = 'tubes successfully found'
+          return response.deserialize.tubes
+        }
       } else {
         this.message = 'there was an error'
         return response
       }
+    },
+    async findSequencescapeTubes () {
+      return this.findTubes(this.sequencescapeTubeRequest)
```
harrietc52: Check if barcodes empty, dont call findTubes

src/views/ScanBarcodes.vue
```diff
@@ -28,31 +30,51 @@ export default {
...
+    },
+    async findTractionTubes () {
+      return this.findTubes(this.tractionTubeRequest)
```
harrietc52: And again, same as above

src/views/ScanBarcodes.vue
```diff
@@ -3,7 +3,9 @@
     <div class="form-group">
       <label for="barcodes">barcodes:</label>
       <textarea type="text" v-model="barcodes" class="form-control" rows="10" cols="10" name="barcodes" id="barcodes" />
-      <b-button id="findTubes" variant="success" @click="findTubes" >Go!</b-button>
+      <b-button id="findSequencescapeTubes" variant="success" @click="findSequencescapeTubes" >find Sequencescape Tubes</b-button>
+      <b-button id="findTractionTubes" variant="success" @click="findTractionTubes" >find Traction Tubes</b-button>
+
```
harrietc52: possibly disable button if no bacodes are entrered

PR #68, opened 2019-03-20
Create child components for runs

stevieing:


harrietc52:
Dependent on https://github.com/sanger/traction-service/pull/36/files

- Doesnt show run name if exists, only run ID
- Doesnt show library barcode if exists
- Possible change library barcode input element to b-form-input

PR #81, opened 2019-04-01
Run buttons to RunMixin

harrietc52:


harrietc52:
I moved the start button to the Runs table @stevieing

tests/unit/views/Runs.spec.js
```diff
@@ -149,19 +86,126 @@ describe('Runs.vue', () => {
...
+      expect(wrapper.find('tbody').findAll('tr').at(0).text()).toMatch(/TRAC-123/)
+    })
+  })
+
+  describe('start button', () => {
+    let button
+
+    it('is enabled when the run state is pending', () => {
+      // run at(3) is in state pending
+      button = wrapper.find('tbody').findAll('tr').at(3).findAll('td').at(4).find('#startRun')
```
stevieing: would it be worth adding an id to each run within the runs table. This would make testing much easier so we could do e.g wrapper.find('#run1').find('#startRun')?

src/mixins/RunMixin.js
```diff
@@ -0,0 +1,103 @@
...
+  name: 'RunMixin',
+  mixins: [Api],
+  data () {
+    return {
+      message: ''
+    }
+  },
+  methods: {
+    async getRun (id) {
+      let promise = await this.runsRequest.find(id)
```
stevieing: Do these promises still need the await? Some have it some don't.

tests/unit/views/Runs.spec.js
harrietc52: Yep sounds good, will do that

src/mixins/RunMixin.js
harrietc52: Ill have a look into it :)

tests/unit/views/Runs.spec.js
harrietc52: Might be a little tricky as the b-table doesnt seem to have anything built in to add id for the row

harrietc52:
@stevieing Happy to merge this in if you are :)

PR #152, opened 2019-08-06
changes to support the pipeline namespace

harrietc52:


harrietc52:
Ok great, maybe add them as issues

PR #158, opened 2019-08-12
Finalise new release method

pjvv:
Linked with https://gitlab.internal.sanger.ac.uk/psd/psd-deployment/merge_requests/14


pjvv:
If this passes the tests it should be good to merge in!

PR #159, opened 2019-08-20
Update table views

pjvv:


src/views/Libraries.vue
```diff
@@ -74,24 +113,95 @@ export default {
...
       }
     },
-    showAlert () {
-      return this.$refs.alert.show(this.message, 'primary')
+    // Get all the libraries
+    // Provider function used my the bootstrap-vue table component
+    async provider() {
+      this.items = await this.getLibraries()
+    },
+    async getLibraries() {
```
stevieing: Is there room for a bit of refactoring here? Libraries and Samples have almost the same method so maybe it could be abstracted out into a mixin. Also the method is quite long. I don't know whether there is any scope for splitting it up a bit?

src/views/Reception.vue
```diff
@@ -77,30 +96,46 @@ export default {
       if (response.successful) {
         this.barcodes = response.deserialize.requests.map(r => r.barcode).join('\n')
       } else {
-        throw 'Failed to create tubes in Traction: ' + response.errors.message
+        throw Error(consts.MESSAGE_ERROR_CREATE_TUBES_FAILED + response.errors.message)
       }
     },
     async handleTractionTubes () {
```
stevieing: Again I am wondering if there is scope for a bit of refactoring? The method is quite lengthy and there seems to be a fair amount of logic checking for the material type.

src/views/Samples.vue
```diff
@@ -78,32 +125,86 @@ export default {
...
+        }
+      } else {
+        throw Error(consts.MESSAGE_ERROR_GET_TRACTION_TUBES)
+      }
+    },
+    async provider() {
+      this.items = await this.getSamples()
+    },
+    // Get all the samples (requests)
+    async getSamples() {
```
stevieing: Same as with the getLibraries() method in the Libraries component.

PR #163, opened 2019-08-22
Added pacbio and saphyr views

stevieing:


src/views/Dashboard.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+    <div class="dashboard">
+      <div class="pipeline saphyr">
+         <router-link :to="{ path: '/saphyr' }" tag="a">
```
harrietc52: Can we remove the description too?

harrietc52:
RunMixin.js

The showRun() function is handling the redirect, but not specifing which pipeline
`this.$router.push({ path: `/run/${runId}` })`

Could be
`this.$router.push({ path: `{{ pipeline }}/run/${runId}` })`

harrietc52:
SaphyrRun.js vue's Back button needs to append the pipeline too
`<router-link :to="{name: 'Runs'}">`

src/views/Dashboard.vue
stevieing: I'm going to leave it there for now. I would like it to be a placeholder so when we come to UAT it gives people a push in the direction of what they will want in there.

PR #174, opened 2019-09-05
Add logging config

pjvv:


src/mixins/Helper.js
```diff
@@ -20,7 +20,7 @@ export default {
      * @param {*} message the message to log
      */
     log(message) {
-      if (process.env !== 'production') {
+      if (process.env.VUE_APP_LOG === 'true') {
```
stevieing: You can either do if (Boolean(process.env.VUE_APP_LOG)) { ... } or if (process.env.VUE_APP_LOG) { ... }

pjvv: Environmental variables are strings and the `Boolean()` function will return `true` for any non-empty string so we have to check another way. https://stackoverflow.com/a/264037

PR #179, opened 2019-09-11
Issue 178 data folder

stevieing:
This allows you to include all of the data files from the test helper rather than having to manually import each one using relative paths.
It would be better if we could import test helper globally in Jest so we no longer need to use relative paths at all.
Obviously the ideal would be to have factories so we don't have to deal with this gubbins at all.


pjvv:
Very eloquent, I like it.

PR #192, opened 2019-09-24
Add pacbio samples

pjvv:


pjvv:
Skipping some tests until discussion about libraries.

PR #200, opened 2019-10-09
Gpl 073 sample sheet generation

harrietc52:
Added Generate Sample Sheet button to PacbioRuns table
Button is a link, simply fetching the sample sheet for a run and handling in the browser
Dependent on traction service PR https://github.com/sanger/traction-service/pull/146
Merge after https://github.com/sanger/traction-ui/pull/199

### DIFF
--- a/src/views/pacbio/PacbioRuns.vue
+++ b/src/views/pacbio/PacbioRuns.vue
@@ -31,6 +31,13 @@
           :per-page="perPage"
           :current-page="currentPage"
           @filtered="onFiltered">
+
+        <template v-slot:cell(actions)="row">
+          <a :id="generateId('generate-sample-sheet', row.item.id)"
+             :href="generateSampleSheetPath(row.item.id)">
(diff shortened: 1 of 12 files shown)


src/views/pacbio/PacbioRuns.vue
```diff
@@ -1,28 +1,124 @@
...
+      try {
+        await this.setRuns()
+      } catch (error) {
+        this.showAlert("Failed to get runs: " + error.message, 'danger')
+      }
+    },
+    generateId(text, id) {
+      return `${text}-${id}`
+    },
+    generateSampleSheetPath(id) {
```
stevieing: I noticed that you put this here rather than adding it to the config. I remember we had a brief discussion about this being difficult to add to the request model. Did you have a look at that? Although we can get away it here I am pretty sure that when we come to tracking for PacBio there will be much more of this. Any thoughts on what we can do?

PR #259, opened 2019-12-18
Gpl 195 drag and drop library into well

stevieing:
All looking good. We both agree this is pretty cool :)


src/components/Well.vue
```diff
@@ -45,14 +50,51 @@ export default {
...
   methods: {
+     ...mapActions([
+    'isLibraryBarcodeValid',
+    'getTubeForBarcode',
+  ]),
+  ...mapMutations([
+    'mutateWell',
+    'addEmptyLibraryToWell',
+    'removeLibraryFromWell',
+    'addLibraryToWell'
```
harrietc52: Think the above mutations aren't needed in this component

src/components/Well.vue
```diff
@@ -45,14 +50,51 @@ export default {
...
   },
   methods: {
+     ...mapActions([
+    'isLibraryBarcodeValid',
+    'getTubeForBarcode',
+  ]),
+  ...mapMutations([
+    'mutateWell',
+    'addEmptyLibraryToWell',
+    'removeLibraryFromWell',
```
harrietc52: Think the above mutations aren't needed in this component

harrietc52:
This is really cool! 
Just added a few small nice to haves. Can do a proper code review once your happy with it

src/components/Well.vue
```diff
@@ -94,4 +136,9 @@ export default {
   .filled {
     fill: purple;
   }
+  // need to sort to add blur
```
harrietc52: What do you mean here?

tests/unit/components/PacbioLibrariesList.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+    it('will have a table', () => {
+        expect(wrapper.find('.libraries')).toBeDefined()
+    })
+
+    it('contains the correct data', () => {
+        expect(wrapper.find('.list-group').findAll('.list-group-item').length).toEqual(5)
+    })
+
+})
```
harrietc52: Could test that the table contains Tube components too

src/components/Tube.vue
```diff
@@ -0,0 +1,71 @@
+<!-- TODO: move inline styles to css -->
```
harrietc52: Still todo?

src/components/Tube.vue
```diff
@@ -0,0 +1,71 @@
...
+      type: String,
+      required: true
+    }
+  },
+  data () {
+    return {
+      isActive: false
+    }
+  },
+  // TODO: need to add a a test for drag
```
harrietc52: Still todo?

src/components/Tube.vue
stevieing: Sort of. I have moved the comment to the tube svg. We should move any styles to css rather than having it inline as it is easier to maintain. We already have this as an issue.

src/components/Tube.vue
stevieing: Yes. I am envisaging that we will start using drag and drop in Saphyr and ONT so we can revisit it then. It definitely needs a test but unsure of how to do it.

src/components/Well.vue
stevieing: It was sorted. Just forgot to remove the comment.

PR #283, opened 2020-01-09
Fix update Run state - reload page

### LINKED ISSUE #270: Update Run state - reload page
Mutators need to be immutable

harrietc52:
Use the vuex pattern:
component calls an action, which does the async call,
then calls a mutator, which updates the state immutably
fixes #270

### DIFF
--- a/src/store/traction/mutations.js
+++ b/src/store/traction/mutations.js
@@ -2,6 +2,12 @@ const mutations = {
     setTags(state, tags) {
         state.tractionTags = tags
     },
+    updateRun(state, run) {
+        let runIndex = state.pacbio.runs.runs.map(e => e.id).indexOf(run.id)
+
+        state.pacbio.runs.runs.splice(runIndex, 1, run)
+        state.pacbio.runs.runs = [...state.pacbio.runs.runs]
+    },
 }
 
 export default mutations
--- a/src/store/traction/actions.js
+++ b/src/store/traction/actions.js
@@ -15,12 +15,17 @@ const cancelRun = async ({ dispatch }, id) => {
     await dispatch('handleRunUpdate', payload)
 }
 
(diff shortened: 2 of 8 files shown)


src/store/traction/mutations.js
```diff
@@ -2,6 +2,12 @@ const mutations = {
     setTags(state, tags) {
         state.tractionTags = tags
     },
+    updateRun(state, run) {
+        let runIndex = state.pacbio.runs.runs.map(e => e.id).indexOf(run.id)
```
stevieing: I was wondering why this is runs.runs?? Not a criticism just looks a bit weird.

src/store/traction/mutations.js
```diff
@@ -2,6 +2,12 @@ const mutations = {
...
+
+        state.pacbio.runs.runs.splice(runIndex, 1, run)
+        state.pacbio.runs.runs = [...state.pacbio.runs.runs]
```
stevieing: I was also wondering what is going on here?

harrietc52: I noticed when I was working on the 'addLibraryToWell' mutation that the speciality feature of vuex, being that it automatically reloads components etc, wasnt working properly and this was becuase we were mutating the vuex state, so the refresh wasnt being trigger. 
The State should be immutable, so the last line there is creating a new object with the old objects state triggering the refresh. So we dont need to recall the provider() functions. 
I think we should try and look throughout the application to make sure we are using Vuex in the vuex way. In that the component calls the action, which then commits a mutation, which updates the state immutably.  


Hope that makes sense!!

src/store/traction/mutations.js
harrietc52: Ye it does! It's just the heirarchy of the state folders, traction/pacbio/runs, then within the runs folder, there is a state called runs.

PR #305, opened 2020-01-13
Converted saphyr and pacbio samples and libraries to use vuex and rem…

stevieing:
…oved the use for mattype.


src/views/saphyr/SaphyrSamples.vue
```diff
@@ -155,7 +152,8 @@ export default {
     },
     async provider() {
       try {
-        this.items = await this.getMaterial(consts.MAT_TYPE_REQUESTS)
+        await this.setRequests()
+        this.items = this.requests
```
harrietc52: We could possibly get rid of items in this component and call this.requests from the b-table, but looks like the clearPreFilter() method is using items. Although not sure how that works, if it does!?

src/views/saphyr/SaphyrLibraries.vue
```diff
@@ -152,7 +151,7 @@ export default {
     // Get all the libraries
     // Provider function used by the bootstrap-vue table component
     async provider() {
-      this.items = await this.getMaterial(consts.MAT_TYPE_LIBRARIES)
+      this.items = await this.setLibraries()
```
harrietc52: Maybe a more vuex'y way is for the provider to use call setLibraries(), then have a separate getter the get the libraries. PacbioRuns.vue does a similar thing

src/views/pacbio/PacbioLibraries.vue
```diff
@@ -143,7 +142,7 @@ export default {
     // Get all the libraries
     // Provider function used by the bootstrap-vue table component
     async provider() {
-      this.items = await this.getMaterial(consts.MAT_TYPE_LIBRARIES)
+      this.items = await this.setLibraries()
```
harrietc52: Same here as below comment

src/store/traction/saphyr/tubes/actions.js
```diff
@@ -71,19 +71,37 @@ const deleteLibraries = async ({ getters }, libraryIds) => {
...
+  let response = await handlePromise(promise)
+  let libraries = null
+
+  if (response.successful && !response.empty) {
+    libraries = response.deserialize.libraries
+    
+    commit('setLibraries', libraries)
+  }
+
+  return libraries
```
harrietc52: may not need to return anything from this function

src/store/traction/saphyr/requests/getters.js
```diff
@@ -0,0 +1,6 @@
+const getters = {
+    requests: state => state.requests,
+    requestsRequest: (state, getters, rootState) => rootState.api.traction.pacbio.requests,
```
harrietc52: could add a small test for requestsRequest

src/store/traction/saphyr/tubes/mutations.js
```diff
@@ -1,6 +1,9 @@
 const mutations = {
   setTubes (state, tubes) {
     state.tractionTubes = tubes
+  },
```
harrietc52: could add small tests for these, don't think any exist at the moment

src/views/saphyr/SaphyrSamples.vue
stevieing: Removed pre filter and modified items.

src/views/saphyr/SaphyrLibraries.vue
stevieing: Done. Also removed prefilter. Removed unneccessary getters.

src/views/pacbio/PacbioLibraries.vue
stevieing: Modified to use getter for libraries. Removed clear pre filter. Removed unneeded getters.

src/store/traction/saphyr/tubes/actions.js
stevieing: I am using the results from the action in the test. Don't know if there is a better way?

src/store/traction/saphyr/requests/getters.js
stevieing: I have added a new issue to review how vuex is setup. I think there are a few getters and other things which are not currently tested.

src/store/traction/saphyr/tubes/mutations.js
stevieing: Again I think we are missing tests for a few mutations.

PR #315, opened 2020-01-17
Added default cost code for pacbio requests.

stevieing:


stevieing:
> All good, would only be if we want this to be in the traction-service. But I'm not sure which one is better :)

Totally agree. I put it in the ui and then though about putting it in the service then got confused so just left it. Time will tell ...

PR #324, opened 2020-01-23
Issue318 wellcolour

BenTopping:


src/components/Well.vue
```diff
@@ -109,6 +109,26 @@ export default {
...
+    },
+    isComplete () {
+      let well = this.well(this.position)
+      if (well === undefined) return false
+      if ((well.movie_time && 
+           well.sequencing_mode && 
+           well.insert_size && 
+           well.on_plate_loading_concentration) !== "" 
+           && well.libraries.length > 0) return true
+      return false
```
harrietc52: Could use the any? and all? type of functions we spoke about

stevieing:
> Maybe we could all regroup about what we want the 'red' well to show.
> 
> It could be 'incomplete' regardless or whether is it missing libraries or metatdata.
> Currently, if a well has metadata but no libraries, then it doesnt show a colour...
> 
> Whereas it could be:
> If the well is complete then green
> Incomplete, red
> Otherwise no colour

Fine by me.
If it has metadata and libraries green
If it has metadata or libraries red
otherwise empty.

PR #330, opened 2020-01-30
Make pacbio run name non editable to prevent clashes

stevieing:


src/components/PacbioRunInfo.vue
```diff
@@ -4,11 +4,11 @@
     <b-row>
       <b-col>
         <b-form-input :value="runName"
-                @change="setName"
                 id="run_name"
                 placeholder="Run name"
                 type="text"
-                width="48"/>
+                width="48"
+                readonly />
```
harrietc52: Nice !

tests/unit/store/traction/pacbio/runs/mutations.spec.js
```diff
@@ -26,10 +26,12 @@ describe('mutateRun', () => {
   })
 
   it('can update the mutate the state, e.g setName', () => {
```
harrietc52: e.g. setSystemName

stevieing:
> All good! The readonly name field on run creation, and then the auto generated name on edit is really nice
> 
> Just a few places to remove run from I think
> 
> PacbioRun.js:
> build()
> createRunPayload()
> updateRunPayload()
> 
> PacbioRunInfo:
> mapMutations setName
> 
> Wasn't sure if in the service we should remove permitting name from the runs controller too?

I wonder if we need to improve the test coverage to ensure that I don't miss those sort of things in the future.

We probably should remove permitting the name from the controller. I will add that as a new issue.

harrietc52:
@stevieing Did the above get removed before merging, build, createRunPayload, updateRunPayload, setName etc?

PR #369, opened 2020-03-16
Refactor of sample metadata modal

BenTopping:


src/views/pacbio/PacbioSamples.vue
```diff
@@ -150,7 +150,7 @@ export default {
   methods: {
     async provider() {
       try {
-        await this.setRequests()
+        this.items = await this.setRequests()
```
harrietc52, line 153: Don't think you need `this.items` as this.setRequests() doesnt return anything
Can also remove items from data

Issue #304, opened 2020-01-13
GPL-322 - As a user I need to have an ONT dashboard


stevieing:

stevieing:
User story
As a long read user when I go to traction I need to see as box which when I click it will take me to all of the ONT data and associated tasks

Who the primary contacts are for this story
Emma Betteridge

Acceptance criteria
To be considered successful the solution must allow:
1 - When I visit the traction home page I will see a view for ONT which will be a portal for accessing ONT tracking

PR #370, opened 2020-03-16
Gpl 322 dashboard refactor

harrietc52:
- Added ONT to Dashboard
- Tried to make Dashboard bit 'prettier'
- Along with refactor, trying to make use of vue components
- Using local storage instead of store to store pipeline state, so it persists over pages. Might be a nicer way to handle localStorage, wanted to prioritise completing the story.


harrietc52:
- I Don't mind, I thought this may feel less clunky and more the default UX behaviour to handle returning back to the home page. Can add 'Dashboard' back in the nav bar if you think?  
- Didn't relise this was a requested feature. Just thought as we add more pipelines the Dashboard would get more clunky with lots of buttons. Again, can add back if you think? 
- Hm not sure what to do about this one. Maybe one for another story to look into how to reduce the UI? For libraries, now the samples are in the extra details, we could remove that column? Or start playing around with the css.

stevieing:
How about this:
- for now add the links back into the boxes
- I will add a story to do some UX work with our customers. Happy to make that a priority as it is becoming an issue and we are second guessing so need some input.
Does that sound reasonable?

harrietc52:
Yep that sounds good, 👍 thanks

harrietc52:
Closes #304

BenTopping:
It is not immediately obvious which pipeline you are looking at when on any given table, maybe adding a header declaring the pipeline would make it clearer. As Steve said, tables spill over (known on chrome) when tables are resized.

harrietc52:
Updated:
- Added the pipeline to all the pipeline views so it is obvious what the user is in
- Made all tables scrollable, horizontally, to help spilling tables and improve UX

harrietc52:
Failing as something in Travis has changed. SS etc effected too

Issue #317, opened 2020-01-20
Naming of fields when editing a PacBio run


stevieing:
When I edit a PacBio run there are no field headings so it is difficult to distinguish. For example when I edit a well it is difficult to tell which field is movie time or insert size. Maybe we could have tooltips or field headings.

PR #393, opened 2020-04-28
GPL-430 - Add titles to pacbio run fields

BenTopping:
Added titles for well modal and tooltips for run info, issue with tooltip warning when tests run


BenTopping:
closes #317

PR #396, opened 2020-04-29
Created GridION SVG

### LINKED ISSUE #296: GPL-326 - As developers we would like to investigate creating an SVG for the ONT Gridion

harrietc52:
Closes #296


src/components/svg/GridionSVG.vue
```diff
@@ -0,0 +1,30 @@
+<template> 
+  <div class="plate">
```
stevieing, line 2: Should this class be something else?

harrietc52:
Good points - thank you

I've renamed the files, GridionSVG => ONTSVG and ONTRun => GridIONRun

ONTSVG  contains the boundary and the slot, the GridIONRun contains the config of what is added to the slot, so here 5 ONTFlowcells are added

A MinionRUN could then be added using the ONTSVG and one flowcell, the Promethion with 24 etc

PR #401, opened 2020-05-04
WIP: Gpl 437 ui pacbio

harrietc52:


harrietc52:
TODO:
1. On Run creation, library isn't being associated with a well
2. Add filter by barcode for library, so can remove validation of a pacbio library in a well as we know it will be a library

Issue #421, opened 2020-05-12
BUG - As a user I need to be able to see more of the plate view


stevieing:
**Describe the bug**
When I currently visit the plates view I can only see a partial view and need to scroll to see the rest.

**To Reproduce**
Steps to reproduce the behaviour:
1. ONT > plates
2. show plate
3. I can only see part of a plate

**Expected behaviour**
I would like to see the full plate

**To Reproduce**
Steps to reproduce the behaviour:
1. ONT > runs > create run
2. Drag and drop tube onto flowcell
3. I can only see part of the name 

When I currently visit the new runs page and add a library I can only see part of the name on the flowcell

When I drag a tube onto the flowcell it would be good if the flowcell is highlighted to see where I am dropping it

**Additional context**
This is a CSS issue with bootstrap.

stevieing:
We can also look at the flowcells as part of this story.

PR #425, opened 2020-05-13
Gpl-378 create ont run

### LINKED ISSUE #297: GPL-378 As a user I need to be able to create an ONT run

harrietc52:
Closes #297


harrietc52:
TODO:
move resolvers into component
move apollo.js file into graphql
reload runs on create
redirect shows alert message
refactor status in flowcell component
check if extra variable is required in gql queries

src/components/ont/OntFlowcell.vue
```diff
@@ -1,10 +1,17 @@
 <template>
-  <g :transform="getMatrix">
-    <rect width="61" height="227"/>
+  <g :transform="getMatrix" v-on:drop="drop" v-on:dragover="allowDrop" >
+    <text x="25" y="30" class="medium">{{ position }}</text>
+
+    <rect width="61" height="227" v-bind:class="status"/>
+
+    <foreignObject y="100" width="70" height="227">
+      <b-form-input v-model="libraryName" placeholder="Name" :id="'libraryNameInput-'+this.position" @change="updateFlowcell"></b-form-input>
```
stevieing, line 8: It is difficult to see the full name of the library when it has been dropped onto the flowcell.

harrietc52: Made it a bit wierd but struggled with SVG HTML/CSS -> Added a tooltip as a quick fix

stevieing: ok. Lets add this issue to #421 . We can do a deep dive into the CSS to see what is going on. If we get some time next week we can have more of a look. I know how much of a pain CSS can be especially when it is wrapped up in bootstrap.

stevieing:
> Is there any chance we could move the apollo.js file to the graphql folder just to keep it all in one place. As we start adding more logic to this it feels like a more logical place to put it.
> 
> UPDATE
> Looked into this, and I think an apollo.js config file is required at the root. I've refactored code of it into the graphql folder

Weird. I tried it in my branch and it was ok. No matter. It is pretty clean anyway.

PR #427, opened 2020-05-14
WIP -Gpl 377

BenTopping:
allows library barcode labels to be printed


harrietc52:
UPDATE: want to be able to print barcodes for a plate too

PR #437, opened 2020-05-19
Footer clean up

BenTopping:
Closes #

* Removed need for RELEASE file
* Changes format of release name data
* ...

### DIFF
--- a/src/components/InfoFooter.vue
+++ b/src/components/InfoFooter.vue
@@ -14,10 +14,9 @@
           </b-col>
           <b-col col lg="3">
             <b-row>Traction [{{ this.environment }}]</b-row>
-            <b-row>{{ this.release }}</b-row>
             <b-row>
               <b-link :href="this.repo">
-                {{ this.repo.slice(38, this.repo.length) }}
+                {{ getRelease() }}
               </b-link>
             </b-row>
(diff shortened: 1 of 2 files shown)


src/components/InfoFooter.vue
```diff
@@ -31,22 +30,25 @@ export default {
...
     return {
         environment: process.env.NODE_ENV,
-        release: "",
         repo: ""
       }
   },
   methods: {
+    getRelease() {
+      if (this.environment != 'development') {
+        return this.repo.slice(51, this.repo.length)
```
harrietc52, line 39: What is 51 and 38? Guessing its number of letters for something? Maybe they could be declared as a constant or data element to make it clean :)

src/components/InfoFooter.vue
```diff
@@ -31,22 +30,25 @@ export default {
...
+      } else {
+        return this.repo.slice(38, this.repo.length)
+      }
+    },
     provider() {
       if (this.environment != 'development') {
         Promise.all([
```
harrietc52, line 46: might be able to use the async/await commands https://dev.to/johnpaulada/synchronous-fetch-with-asyncawait  instead of Promise.all

PR #439, opened 2020-05-20
Gpl 381 edit runs

### LINKED ISSUE #303: GPL-381 As a user I want to be able to edit ONT GridION runs
User story
As a user I need to be able to edit an ONT run

Who the primary contacts are for this story
Steve / Harry

Acceptance criteria

    I should be able to modify the state of a run
    I should be able to modify which libraries are in which flowcells

Additional context

harrietc52:
Closes #303 
 
TODO:
~- tests~
~refactor out resolvers and schema~
~- Rename getCovidRun to getOntRun (check ontRun vs covidRun with service - maybe choose one?)~
~- check run state~
~- change name of getClientFlowcell to reflect that it returns a libraryName~
~- check what returns are needed on client mutations~
~- refactor create variales in run compoennt with js destructing~


src/views/ont/OntHeronRun.vue
```diff
@@ -56,70 +80,81 @@ export default {
...
+        }))
+
+      let variables = {
+        flowcells: flowcells
+      }
+
+      if (!this.newRecord) {
+        variables.id = this.id
+      }
+
```
harrietc52, line 110: Is there a nicer way to do the above?

PR #450, opened 2020-05-22
Rename endpoints for ONT mutations

sdjm-tessella:
Confirmed as working by Me, Chris and Steve


src/graphql/queries/CreatePlateWithSamples.mutation.gql
```diff
@@ -1,5 +1,5 @@
 mutation ($barcode: String!, $wells: [WellWithSampleInput!]!) {
-  createPlateWithCovidSamples( 
+  createPlateWithSamples(
```
Chris-Friend, line 2: We've gone for `createPlateWithSamples` here...

src/graphql/queries/PoolSamples.mutation.gql
```diff
@@ -1,5 +1,5 @@
 mutation ($plate_barcode: String!) {
-  createCovidLibraries (
+  createOntLibraries (
```
Chris-Friend, line 2: ...but `createOntLibraries` here. I'm wondering why these are different? Both creations are in the Ont pipeline

sdjm-tessella: It was a decision I made based on the inputs.  While plates are not Ont specific in themselves and you can pass in an ambiguous set of inputs, the OntLibraries endpoint needs you to tell it which plate you're going to create libraries for and that plate needs to have samples of a specific type on them.  It was somewhat arbitrary, but it felt like in the former mutation you should specify which type of samples you want and there's no way to confirm which type you sent from just the input hash.  In the latter case, you can confirm that the plate is of the right type.

Chris-Friend: And that is why the GraphQL create plates mutation now includes a study type argument?

Issue #448, opened 2020-05-22
GPL-491 When a library is created I would like to see a more informative success message


stevieing:
**User story**
When a library is created I would like to see a more informative success message

**Who the primary contacts are for this story**
Steve / Harry

**Acceptance criteria**
To be considered successful the solution must allow:
1. When a library is created I would like to see a message which includes the library name

**Additional context**

PR #452, opened 2020-05-26
Gpl nnn library creation message

Chris-Friend:
Links to #448 

* Update the library creation request to return library names 
* Show the library names rather than the tube barcodes on successfully pooling samples
* Fix existing broken test

### DIFF
--- a/src/components/ont/OntPlate.vue
+++ b/src/components/ont/OntPlate.vue
@@ -52,7 +52,7 @@ export default {
         if (response.errors.length > 0) {
           this.$emit('alert', 'Failure: ' + data.data.createOntLibraries.errors.join(', '), 'danger')
         } else {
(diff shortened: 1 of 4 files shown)


src/components/ont/OntPlate.vue
```diff
@@ -52,8 +52,8 @@ export default {
         if (response.errors.length > 0) {
           this.$emit('alert', 'Failure: ' + data.data.createOntLibraries.errors.join(', '), 'danger')
         } else {
-          let libraryBarcodes = response.tubes.map(t => t.barcode).join(', ')
-          this.$emit('alert', 'Library(s) were created with barcodes: ' + libraryBarcodes, 'success')
+          let libraryNames = response.tubes.map(t => t.materials.map(m => m.name).join(', ')).join(', ')
```
sdjm-tessella, line 55: I assume you could `flat_map` here and only `.join` the once?

PR #454, opened 2020-05-26
When a run is created stay on the same page with a success message.

### LINKED ISSUE #444: GPL-492 When I click the create run button I would rather stay on the page
**User story**
When I click the create run button I would rather stay on the page than be taken back to the run page

**Who the primary contacts are for this story**
Steve / Harry

**Acceptance criteria**
1. When I click on the create run button and it is successful I would like to see a success message
2. When I click on the create run button and it is successful I would like to stay on the page and be able to edit the run

**Additional context**

stevieing:
Closes #444

* when a run is created stay on the run page
* Show a success message

TODO:
* when a run is created add it to the cache?

### DIFF
--- a/src/views/ont/OntHeronRun.vue
+++ b/src/views/ont/OntHeronRun.vue
@@ -88,7 +88,7 @@ export default {
         if (response.errors.length > 0) {
           this.showAlert('Failure: ' + response.errors.join(', '), 'danger')
         } else {
-          this.redirectToRuns()
+          this.showAlert('Run successfully created', 'success')
         }
       })
     },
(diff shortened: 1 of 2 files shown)


src/views/ont/OntHeronRun.vue
```diff
@@ -88,7 +88,7 @@ export default {
         if (response.errors.length > 0) {
           this.showAlert('Failure: ' + response.errors.join(', '), 'danger')
         } else {
-          this.redirectToRuns()
+          this.showAlert('Run successfully created', 'success')
```
sdjm-tessella, line 91: Do you also need to update the `run` variable based on the response?

stevieing: Yes. Thought I would wait until the caching issue was sorted because I guess I would be using the same code to update the run variable.

harrietc52:
It could get the run ID of the recenently created run, and redirect to the newly created run page 
As this is the same component, you could still show the alert then and the create button would automatically then become update button

PR #455, opened 2020-05-26
Prevent Dialog Box

### LINKED ISSUE #443: BUG - Sometimes when I drag a library on the create run page it opens a dialog box
**Describe the bug**
If I drag a libray and select the tube picture it opens a dialog box asking me to save the picture

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page ONT > Runs > New
2. Select and dag tube
3. Dialog box opens asking me if I want to save the picture

**Expected behaviour**
Nothing

**Screenshots**

* Windows or Mac?

**Additional context**

Chris-Friend:
Closes #443 

* Prevent the default drop action so in Firefox the tube image is not downloaded

### DIFF
--- a/src/components/ont/OntTube.vue
+++ b/src/components/ont/OntTube.vue
@@ -7,7 +7,7 @@
     v-on:mouseover="isActive = true" 
     v-on:mouseleave="isActive = false"
   >
-    <b-img left src="/tube.png" height="30" />
+    <b-img left src="/tube.png" height="30" draggable="false" />
     <div class="name">
       {{ name }}
     </div>
--- a/src/components/ont/OntTube.vue
+++ b/src/components/ont/OntTube.vue
@@ -16,6 +16,9 @@
 
 <script>
 
+const img = new Image()
+img.src = '/tube.png'
+
 export default {
   name: 'OntTube',
   props: {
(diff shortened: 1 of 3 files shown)


src/components/ont/OntTube.vue
```diff
@@ -16,6 +16,9 @@
 
 <script>
 
+const img = new Image()
+img.src = '/tube.png'
+
```
sdjm-tessella, line 21: What did changing the scope of this constant achieve?  Making it truly constant?

Chris-Friend: I noticed that if I didn't call `setDragImage` the drag was working as expected in Safari. After some digging I found [this](https://stackoverflow.com/a/22610927) slightly outdated StackOverflow, which did the trick.

I also noticed that if you kept defining the image in the event handler, on Safari it got added as a Source file each time (not sure if just a reference to the same image, a browser cache, or something else).

Pre-defining it and increasing it's scope fixed both of these issues

Chris-Friend: As per the context of the change: the drag wasn't working in Safari

stevieing: Strangely enough it was already like this in pacbio tube. As we are doing the same thing.

Chris-Friend: Could've saved myself some time if I'd looked at that 😆

stevieing: My fault. Should have pointed you to it. I assumed that the same drag and drop had been used. We need to abstract it into the same place as they do pretty much exactly the same thing.

PR #469, opened 2020-06-03
Added spinner to OntReception and minor refactor

### LINKED ISSUE #466: GPL-518 As a developer I would like to add feedback to the reception screen when plates are being imported to let users know that something is happening
**User story**
As a developer I would like to add feedback to the reception screen when plates are being imported to let users know that something is happening

Steve / Harry

**Acceptance criteria**
- when plates are being loaded there should be a loading spinner
- disable the import button so users don't try and press it again

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

harrietc52:
Closes #466

### DIFF
--- a/src/views/ont/OntReception.vue
+++ b/src/views/ont/OntReception.vue
@@ -49,7 +49,7 @@ export default {
   },
   methods: {
     async getSequencescapePlates() {
-      await getPlates(this.sequencescapeRequest, this.formattedBarcodes)
+      return await getPlates(this.sequencescapeRequest, this.formattedBarcodes)
     },
     async handleCreateTractionPlates () {
       this.busy = true
--- a/src/views/ont/OntReception.vue
+++ b/src/views/ont/OntReception.vue
@@ -2,6 +2,10 @@
   <div class="reception">
     <alert ref='alert'></alert>
 
+    <b-modal v-model="busy" hide-footer hide-header no-close-on-backdrop>
+      <spinner size="huge" message="Importing plates..."></spinner>
+    </b-modal>
+
     <div class="form-group">
       <label for="barcodes">Barcodes:</label>
       <textarea type="text"
@@ -15,8 +19,8 @@
     <b-button class="scanButton"
               id="createTractionPlates"
               variant="success"
-              @click="createTractionPlates"
-              :disabled="this.barcodes.length === 0">
+              @click="handleCreateTractionPlates"
+              :disabled="isDisabled">
       Import
     </b-button>
   </div>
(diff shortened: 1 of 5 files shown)


src/views/ont/OntReception.vue
```diff
@@ -27,50 +31,53 @@ import Alert from '@/components/Alert'
...
+    Spinner
   },
   methods: {
-    async getSequencescapePlates (barcodes) {
-      return await getPlates(this.$store.getters.api.sequencescape.plates, barcodes)
-    },
-    getBarcodes () {
-      return this.barcodes.split('\n').filter(Boolean).join(',')
+    async getSequencescapePlates() {
+      await getPlates(this.sequencescapeRequest, this.formattedBarcodes)
```
stevieing, line 52: I think this needs a return

src/views/ont/OntReception.vue
```diff
@@ -27,50 +31,53 @@ import Alert from '@/components/Alert'
...
     },
-    async handleSequencesapePlates () {
-      let jsonPlates = await this.getSequencescapePlates(this.getBarcodes())
-      if (jsonPlates !== undefined) {
-        this.plates = transformPlates(jsonPlates)
+    async handleCreateTractionPlates () {
+      this.busy = true
+      let ssPlatesJson = await this.getSequencescapePlates()
+
+      if (ssPlatesJson.length === 0) {
```
stevieing, line 58: I think this should be 
`if (ssPlatesJson.length === undefined) {`

src/views/ont/OntReception.vue
harrietc52: Oh interesting thanks, I couldnt test with valid barcodes becuase the plate I was generating didnt have tagged samples, so might need you to test again after changes! But ill add that return :)

src/views/ont/OntReception.vue
stevieing: Is it possible to add a test for it? You could add a test to say that when ssPlatesJson is undefined it sets busy to false.

src/views/ont/OntReception.vue
stevieing: Again could we add a test for this?

harrietc52: Oh Its a method, not a computed, so dont think it needs the return?

src/views/ont/OntReception.vue
harrietc52: Adding the return to the above returns an empty array, so will keep the condition the same

src/views/ont/OntReception.vue
harrietc52: Ah I remember trying, but struggled as getPlates is being imported in from the SequencescapePlates api, so couldnt mock it.

harrietc52: Ah I remember now trying, but couldnt as getPlates is imported so struggled to mock it

Issue #467, opened 2020-06-02
GPL-526 UI Pagination


stevieing:
**User story**
As a user I would like the libraries in the run to be ordered by most recent first as we would generally want to use ones which have most recently been created

Pagination for Plates/ Libraries/ Runs/ LibrariesList

**Who are the primary contacts for this story**
Steve / Harry

**Acceptance criteria**
To be considered successful the solution must allow:
- When I go to the create run page the most recent libraries should appear first
- [ ] ...

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

harrietc52:
Coupled with https://github.com/sanger/traction-service/issues/409

PR #478, opened 2020-06-12
Graphql pagination UI

harrietc52:
Closes https://github.com/sanger/traction-ui/issues/467
Coupled with https://github.com/sanger/traction-service/pull/449

### DIFF
--- a/src/components/ont/OntFlowcell.vue
+++ b/src/components/ont/OntFlowcell.vue
@@ -39,6 +39,7 @@ export default {
     }
   },
   methods: {
+    // Update is being called twice when a library is dragged?
     updateFlowcell (libraryName) {
       this.$emit('updateFlowcell', this.position, libraryName)
     },
--- a/src/components/ont/OntFlowcell.vue
+++ b/src/components/ont/OntFlowcell.vue
@@ -43,15 +43,21 @@ export default {
       event.dataTransfer.setData('flowcellPosition', this.position)
       event.dataTransfer.setData('libraryName', this.library.name)
 
+      event.dataTransfer.setData('sourceType', 'FLOWCELL')
+
       this.hover = false
     },
     drop (event) {
       event.preventDefault()
       let libraryName = event.dataTransfer.getData('libraryName')
-
       this.updateFlowcell(this.position, libraryName)
-      this.updateLibraryList(libraryName, true)
 
+      if (event.dataTransfer.getData('sourceType') === 'FLOWCELL') {
+        let flowcellPosition = parseInt(event.dataTransfer.getData('flowcellPosition'))
+        this.updateFlowcell(flowcellPosition, '')
+      } else if (event.dataTransfer.getData('sourceType') === 'LIBRARY_LIST'){
+        this.updateLibraryList(libraryName, true)
+      }
       this.hover = false
     },
   },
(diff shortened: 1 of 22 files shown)


harrietc52:
TODO:
- check drag and drop from one flowcell to another flowcell
- `TODO` test in `OntRunLibrariesList`

src/components/ont/OntRunLibrariesList.vue
```diff
@@ -12,58 +12,66 @@
...
+
       this.hover = false
     },
-    isLibrarySelected(library) {
-      return this.selectedLibraryNames.includes(library.name)
+    fetchLibraries () { 
+      this.$apollo.query({
+        query: LIBRARIES_ALL_QUERY,
+        variables: {
+          unassignedToFlowcells: false,
```
stevieing, line 47: Should this be true?

src/views/ont/OntLibraries.vue
```diff
@@ -52,31 +62,36 @@ export default {
...
   },
   methods: {
+    handlePrint(printer) {
+      this.handlePrintLabel(printer)
+    },
+    getLibraries(ctx, callback) {
+      this.$apollo.query({
+        query: LIBRARIES_ALL_QUERY,
+        variables: {
+          unassignedToFlowcells: false,
```
stevieing, line 80: Should this also be true? Unless I'm misunderstanding.

src/views/ont/OntPlates.vue
```diff
@@ -38,37 +47,40 @@ export default {
...
-    refetchPlates() {
-      this.$apollo.queries.plates.refetch()
+    getPlates(ctx, callback) {
+      this.$apollo.query({
+        query: PLATES_ALL_QUERY,
+        variables: {
+          pageNum: ctx.currentPage,
+          pageSize: ctx.perPage
+        },
+        fetchPolicy: 'no-cache'
```
stevieing, line 72: More a query than a code review comment. Would we be able to cache this in future?

src/components/ont/OntFlowcell.vue
```diff
@@ -32,36 +34,32 @@ export default {
...
+
       this.hover = false
     },
     drop (event) {
       event.preventDefault()
-      this.updateFlowcell(event.dataTransfer.getData('name'))
+      let libraryName = event.dataTransfer.getData('libraryName')
+      this.updateFlowcell(this.position, libraryName)
+
+      if (event.dataTransfer.getData('sourceType') === 'FLOWCELL') {
```
stevieing, line 55: why do you have to do this?

harrietc52: It's for drag and drop from one flowcell to another. Depending if the flowcell got dragged from the library list/ flowcell, it needs to update different things

src/components/ont/OntRunLibrariesList.vue
harrietc52: Just to add unassignedToFlowcells: false, means get all libraries

src/views/ont/OntPlates.vue
harrietc52: Yes I think so

src/views/ont/OntLibraries.vue
harrietc52: Its getting all libraries, including a assignedToFlowcell field, which is being displayed in the front end table. This could be filtered on in the future

PR #540, opened 2020-10-26
Gpl 686 modify pacbio tags

### LINKED ISSUE #533: GPL-686 As a long read user I would like to be able to modify library tags without an RT request
**User story**
As a long read user I would like to be able to modify library tags without an RT request

Craig

**Acceptance criteria**
- [ ] when I edit a library I should be able to change a tag

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #533 

*You can now edit tags within the show details button on pacbio libraries
*
* ...

### DIFF
--- a/src/components/pacbio/PacbioTagEditModal.vue
+++ b/src/components/pacbio/PacbioTagEditModal.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <b-button :id="'editTag'" size="sm" @click="show" variant="outline-primary">Edit</b-button>
+    <b-button :id="'editTag'+this.tag.id" size="sm" @click="show" variant="outline-primary">Edit</b-button>
 
     <b-modal
       id="editTagModal"
@@ -61,9 +61,10 @@ export default {
         await this.updateTag(payload)
         this.alert('Tag updated', 'success')
       } catch (err) {
-        this.alert('Failed to update tag. ' + err, 'danger')
+        this.alert('Failed to update Tag. ' + err, 'danger')
       }
       this.hide()
+      this.$emit('reloadPage')
     },
     async provider() {
       try{
--- a/src/components/pacbio/PacbioTagEditModal.vue
+++ b/src/components/pacbio/PacbioTagEditModal.vue
@@ -66,14 +66,14 @@ export default {
     },
     async provider() {
       try{
-         await this.setTags()
-         this.tags = this.tractionTags
-            .filter(tag => tag.tag_set_id == 1)
-            .map(tag => tag.group_id)
+        await this.setTags()
+        this.tags = this.tractionTags
+          .filter(tag => tag.tag_set_id == 1)
+          .map(tag => ({ text: tag.group_id, value: tag.id }))
       } catch (error) {
-        this.showAlert("Failed to get libraries or tags: " + error.message, 'danger')
+        this.alert("Failed to get tags: " + error.message, 'danger')
       }
-            console.log(this.tags)
+      console.log(this.tags)
     },
     show() {
       this.$refs['modal'].show()
--- a/src/components/pacbio/PacbioTagEditModal.vue
+++ b/src/components/pacbio/PacbioTagEditModal.vue
@@ -10,12 +10,12 @@
 
       <b-form id="editTagForm" v-on:submit.prevent>
         <b-form-group id="input-group-1"
-                      :label="'Sample tag: '"
+                      :label="'Sample tag: ' + this.tag.sample_name"
                       label-for="input-1">
           <b-form-select ref="tagSelection" 
                         id="tagSelection" 
-                        :value="123"
-                        :options=[]>
+                        :v-model="selectedTagSampleId"
+                        :options="tags">
           </b-form-select>
         </b-form-group>
       </b-form>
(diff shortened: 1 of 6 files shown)


src/views/pacbio/PacbioLibraries.vue
```diff
@@ -66,6 +66,10 @@
               :items= row.item.requests
               :fields="field_in_details"
               :filter="filter">
+                <template v-slot:cell(edit_tag)="row">
+                  <PacbioTagEditModal :tag="row.item" @alert="showAlert" @reloadPage="provider" >
```
harrietc52, line 70: Ah haven't seen the `@reloadPage` before - that's nice!

src/views/pacbio/PacbioLibraries.vue
```diff
@@ -118,7 +123,8 @@ export default {
   components: {
     Alert,
     PrinterModal,
-    PacbioLibraryEditModal
+    PacbioLibraryEditModal,
+    PacbioTagEditModal
```
harrietc52, line 127: Neat 👍

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+  name: 'PacbioLibraryTagModal',
+  mixins: [ModalHelper],
+  data() {
+    return {
+      tags: [],
+      selectedSampleTagId: ""
+    }
+  },
+  props: {
+    tag: {
```
harrietc52, line 53: Renaming this :)

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+  <div>
+    <b-button :id="'editTag'+this.tag.id" size="sm" @click="show" variant="outline-primary">Edit</b-button>
+
+    <b-modal
+      id="editTagModal"
+      ref="modal"
+      title="Edit Tag"      
+    >    
+
+      <b-form id="editTagForm" v-on:submit.prevent>
```
harrietc52, line 11: What does submit.prevent do?

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+        <b-form-group id="input-group-1"
+                      label="Sample tag: "
+                      label-for="input-1">
+                      <span>{{ this.tag.sample_name }}</span>
```
harrietc52, line 15: Dont think you need span and maybe the sample name could go along side Sample tag so its would read in one line

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+  },
+  props: {
+    tag: {
+      type: [Object]
+    }
+  },
+  methods: {
+    async update() {
+      try {
+        let payload = {selectedSampleTagId: this.selectedSampleTagId, request_library_id: this.tag.id }
```
harrietc52, line 60: this.tag will be updated too from the rename :)

harrietc52: just check over your spaces and formatting throughout

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+        await this.updateTag(payload)
+        this.alert('Tag updated', 'success')
+      } catch (err) {
```
harrietc52, line 63: I'm not sure this ever gets hit? Does `updateTag` throw an exception that can be caught or do you have to check in the response and then handle the alert message accordingly?

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+        this.alert('Failed to update Tag. ' + err, 'danger')
+      }
+      this.hide()
+      this.$emit('reloadPage')
```
harrietc52, line 67: Hm this makes me wonder that something isn't quite right if you have to manually reload the page - might be something in the VueX store asyn reloading. Will have a proper look after lunch

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+    },
+    async provider() {
+      try{
+        await this.setTags()
+        this.tags = this.tractionTags
+          .filter(tag => tag.tag_set_id == 1)
```
harrietc52, line 73: How come its tag_set_id==1? If its important probably good form to make it a constant, or add a clean comment etc

src/components/pacbio/PacbioTagEditModal.vue
```diff
@@ -0,0 +1,102 @@
...
+          .map(tag => ({ text: tag.group_id, value: tag.id }))
+      } catch (error) {
```
harrietc52, line 75: Again just want to check that an error is throw in setTags that could be caught here

harrietc52 (commented):
Added a few comments, its really not far off though :) any questions let me know 

Also, I can't see how it wouldn't work if a library did happen to have more than one sample and tag, as the modal is for each row of the sample/tag lists, but maybe worth checking with a dummy bit of data to check. I know this won't be the case anyway but think the code could handle it if it was

src/components/pacbio/PacbioTagEditModal.vue
BenTopping: not needed in this instance, I will remove it, it is a vue method with stops a form being submitted

src/components/pacbio/PacbioTagEditModal.vue
BenTopping: OK, changed so it reads Sample tag: {sample_name}

src/components/pacbio/PacbioTagEditModal.vue
BenTopping: The tag_set_id == 1 is the pacbio tag set, other tag_sets such as ONT may have tag_set_id == 2, this is done so not every tag in traction is displayed in the drop down, only the pacbio ones

harrietc52: Ok :) So maybe just adding a comment or making the 1 a constant with name PACBIO_TAG_SET to make it clear

harrietc52 (approved):

PR #564, opened 2021-01-06
incremented release version

harrietc52:


stevieing (approved):
How often do we caught out on this one!

PR #595, opened 2021-05-13
Added Cypress test.

stevieing:
If you already have traction-ui set up you will need to remove node_modules folder and rebuild using yarn.

You will need to make the following changes to get it working:
- if you don't already have one create .env.production.local in the root folder as a copy of .env.production
- change `VUE_APP_TRACTION_BASE_URL=REPLACE_VUE_APP_TRACTION_BASE_URL` to `VUE_APP_TRACTION_BASE_URL=http://traction`
- change `VUE_APP_SAMPLEEXTRACTION_BASE_URL=REPLACE_VUE_APP_SAMPLEEXTRACTION_BASE_URL` to `VUE_APP_TRACTION_BASE_URL=http://samples-extraction`

To run:

- `yarn test:e2e`
- Then when the cypress window opens click on import_samples_from_samples_extraction.js

If successfull you should see something like the following:
Screen Shot 2021-05-13 at 2 19 05 PM

Thoughts and conclusions:
- It was easy to set up
- The syntax is simple and readable
- Mocking apis is a breeze using intercept
- Mocking responses is a breeze using fixtures
- It is much easier than writing Jest unit tests
- No dependencies as Cypress is installed within your app.
- You can remove the mocks to use it for testing locally against running servers


.gitignore
```diff
@@ -33,3 +33,5 @@ coverage
 
 # talisman
 .talismanrc
+
+.DS_Store
```
pjvv, line 37: It's helpful have `.DS_Store` in a global `.gitignore` file: https://stackoverflow.com/a/7335487

.release-version
```diff
@@ -1 +1 @@
-3.4.0
+3.4.0
```
pjvv, line 1: This hasn't been bumped but the last line has been removed?

package-lock.json
```diff
@@ -0,0 +1,19003 @@
+{
```
pjvv, line 1: We are using yarn so we shouldn't have `package-lock.json`, only `yarn.lock`.

BenTopping (approved):
Looks good! Maybe also put your comment on the PR in the readme (about env.production.local) , the issue with the footer having random text is caused by a fetch to a non existent file in the infoFooter component because the environment is production. Is there a reason the environment needs to be production?

README.md
```diff
@@ -55,7 +55,15 @@ To run the app and have hot-reloads for development:
 
 - Running end to end tests:
 
-      yarn run test:e2e
+    yarn run test:e2e
+
+    You will need to make the following changes to get it working:
+        * if you don't already have one create `.env.production.local` in the root folder as a copy of `.env.production`
+        * change `VUE_APP_TRACTION_BASE_URL=REPLACE_VUE_APP_TRACTION_BASE_URL to VUE_APP_TRACTION_BASE_URL=http://traction`
+        * change `VUE_APP_SAMPLEEXTRACTION_BASE_URL=REPLACE_VUE_APP_SAMPLEEXTRACTION_BASE_URL to VUE_APP_TRACTION_BASE_URL=http://samples-extraction`
```
BenTopping: 'to VUE_APP_TRACTION_BASE_URL=http://samples-extraction' should be VUE_APP_SAMPLEEXTRACTION_BASE_URL=http://samples-extraction ?

PR #599, opened 2021-05-21
Gpl 281 import samples from sequencescape

### LINKED ISSUE #336: GPL-281 As any user I should be able to import samples into traction directly from Sequencescape
**User story**
As any user I should be able to import samples into traction directly from Sequencescape or Samples extraction

Craig / Steve

**Acceptance criteria**
- [ ] When I enter a Sequencescape barcode in the pacbio reception it will pull back the correct information for that plate
- [ ] When I enter a Samples Extraction barcode it will pull back the correct information for that tube
- [ ] When I submit the information it will create a plate or tube which can be movec to library preparation

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

stevieing:
Closes #336 

* created end points to create a pacbio plate in traction
* created a pacbio reception page

Still to do:
* add view to pacbio home page
* create different method to transform plates using snake case rather than camel case.


src/services/Sequencescape.js
```diff
@@ -0,0 +1,85 @@
...
+*/
+const OntSample = (aliquot) => ({
+  externalId: aliquot.sample.uuid,
+  name: aliquot.sample.name,
+  tagOligo: aliquot.tag_oligo,
+  externalStudyId: aliquot.study.uuid,
+})
+
+/*
+  This is in snake case as the end point is JSON API
```
JamesGlover, line 62: I noticed the 1.1 version of the json API spec is now recommending camel case.

src/services/Sequencescape.js
```diff
@@ -0,0 +1,85 @@
...
+  * name - the sample name
+  * tagOligo - the oligo of the aliquot
+  * external study id - the study uuid of the sample
+*/
+const PacbioSample = (aliquot) => ({
+  external_id: aliquot.sample.uuid,
+  name: aliquot.sample.name,
+  external_study_id: aliquot.study.uuid,
+  species: aliquot.sample.sample_metadata.sample_common_name,
+})
```
JamesGlover, line 74: I can't quite decide if these belong here or not, it feels like they might be better off in their respective pipelines, as they are mostly concerned with the payload sent back to traction, and are obviously couple to the pipeline itself.

On the flip-side however they are also dependent on Seqeuncescape's implementation of aliquot.

JamesGlover (approved):
All looks good, although my JS and Vue is a little rusty.

One thing I did notice, is that the API calls aren't using select at all. Depending on performance tis may be worth considering for SS plates, as some attributes, like state, are a bit expensive.

src/api/Config.json
```diff
@@ -1,119 +1,121 @@
...
+                            },
+                            {
+                                "name": "wells",
+                                "include": ""
+                            }
+                        ]
+                    },
+                    {
+                        "name": "plates",
+                        "include": "plates.wells.materials"
```
BenTopping, line 79: this can be just include:"wells.materials" but that change is also in my PR so ill just wait til merge and resolve the conflict

.release-version
```diff
@@ -1 +1 @@
-3.6.2
\ No newline at end of file
+3.7.2
```
BenTopping, line 1: when bumping by minor the patch should be set back to 0

BenTopping (dismissed):

src/services/Sequencescape.js
stevieing: Agreed. Still not sure.

PR #600, opened 2021-05-24
GPL-878-2 Add plates view to Pacbio

BenTopping:
Closes #

* Added plates view to pacbio pipeline
* Added plate info modal as placeholder for adding sample pooling / metadata after clarification with long read
* Edited name of ONT plate query to be accurate


tests/e2e/specs/pacbio_plates_view.js
```diff
@@ -0,0 +1,16 @@
+describe('Pacbio plates view', () => {
```
stevieing, line 1: It would also be worth testing that the plate is there and has a filled well.

stevieing (changes requested):
I think just make an addition to the e2e test. I suspect there will be some conflicts once I have merged my plate reception stuff.

tests/e2e/specs/pacbio_plates_view.js
BenTopping: 👍

JamesGlover (approved):

PR #601, opened 2021-05-25
Fixed bug for runs/plates and runs/wells.

### LINKED ISSUE #598: DPL-022 BUG - Because we are now creating pacbio plates the end points for creating sequencing plates now no longer work.
**Describe the bug**
Because we are now creating pacbio plates the end points for creating sequencing plates now no longer work.
Creating pacbio runs used the pacbio/plates and pacbio/wells endpoints. These have now been changed to pacbio/runs/plates and pacbio/runs/wells.
We need to fix the ui so it points to these endpoints.
This will require changing the config.json and how requests are constructed.

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**
Screen Shot 2021-05-20 at 4 33 38 PM

* Windows or Mac?

**Additional context**

stevieing:
Closes #598 

*
*
* ...


src/api/ApiBuilder.js
```diff
@@ -1,51 +1,101 @@
...
+ * @param {*} config - A piece of json containing all the apis, resources and pipeline
+ * @param {*} environment - The current node environment which should contain the variables for the base URLs which are sensitive
+ * @returns {*} an object of apis
+ */
+const build = ({ config, environment }) => {
+  return config.reduce((result, api) => {
+    return {
+      ...result,
+      // each api will be a property and we need to extract the baseURL from the environment
+      [api.name]: buildApi({ ...api, baseURL: environment[`${api.baseURL}`] }),
```
JamesGlover, line 29: Is there a reason for the interpolation here? Would:

```js
environment[api.baseURL]
```

Do something different?

tests/unit/api/PacbioRun.spec.js
```diff
@@ -121,80 +123,81 @@ describe('Run', () => {
...
 
-      api = build(Api.Config, process.env)
-      api.traction.pacbio.runs.create = jest.fn()
-      api.traction.pacbio.plates.create = jest.fn()
-      api.traction.pacbio.wells.create = jest.fn()
-      api.traction.pacbio.runs.destroy = jest.fn()
-      api.traction.pacbio.plates.destroy = jest.fn()
-      api.traction.pacbio.wells.destroy = jest.fn()
+      api = build({ config: Api.Config, environment: process.env })
+      pacbioRequest = api.traction.pacbio
```
JamesGlover, line 127: Aww, thought you could dry this out further to just `api.traction.pacbio.runs`, but nope, `api.traction.pacbio` sneaks in at the bottom on its own.

JamesGlover (dismissed):

src/api/ApiBuilder.js
stevieing: Good point. It is probably just an artifact from the previous code.

tests/unit/api/PacbioRun.spec.js
stevieing: Exactly. I will come back to this later anyway. I have seen some great advice recently on testability which would mean you can create a wrapper around the original api with a proxy so you don't have to mock out the individual calls but that is something for another day.

PR #603, opened 2021-05-28
Issue589/gpl 992 as a long read user i would like to be able to map a library back to the original samples

JamesGlover:
Closes #

*
*
* ...


stevieing (approved):
Just need to bump the release.

PR #605, opened 2021-06-07
Bug -wells would be invisible if they had no material

BenTopping:
wells would be invisible if they had no material on the pacbio plates page, suspect caused by changes in backend during development of initial feature.

Closes #

*
*
* ...


src/components/plates/PlateItem.vue
```diff
@@ -53,7 +53,7 @@ export default {
   methods: {
     getWellAt(position) {
       let well = this.wells.filter((well) => well.position == position)[0]
-      return well ? well : { position: position, materials: [] }
+      return well && well.materials !== undefined ? well : { position: position, materials: [] }
```
stevieing, line 56: Can you just do `well && well.materials`?
Also you can do `{ position, materials: [] }`
In javaScript if the property name equates to an existing variable it is inferred i.e. within an object position is the same as position: position.

stevieing (approved):
Just a couple of minor JavaScript suggestions.

PR #618, opened 2021-06-14
pacbio request cost code default

BenTopping:
Closes #

* When importing a tube via samples extraction cost code is now removed from request and defaulted in back end if left blank
*
* ...


src/store/traction/pacbio/requests/actions.js
```diff
@@ -87,7 +76,7 @@ const sampleExtractionTubeJson = (tubes) => {
         library_type,
         estimate_of_gb_required,
         number_of_smrt_cells,
-        cost_code: processCostCode(t),
+        ...(t.cost_code === null ? null : { cost_code: t.cost_code }),
```
stevieing, line 79: I think all you need to do is make cost_code part of the mapping i.e. 

```
return tubes.map(
    ({
      library_type,
      estimate_of_gb_required,
      number_of_smrt_cells,
      barcode,
      study_uuid: external_study_id,
      sample_uuid: external_id,
      cost_code,
      fields: { sanger_sample_id: name, sample_common_name: species },
      ...t
    }) => ({
      sample: { name, species, external_id },
      request: {
        external_study_id,
        library_type,
        estimate_of_gb_required,
        number_of_smrt_cells,
        cost_code,
      },
      tube: { barcode },
    }),
  )
```

tests/unit/store/traction/pacbio/requests/actions.spec.js
```diff
@@ -127,39 +127,13 @@ describe('actions', () => {
...
 
-    it('if not completed and no DTOL sample leave blank', () => {
-      tube.cost_code = null
-      expect(Actions.processCostCode(tube)).toBeNull()
+    it('if cost code is null do not include cost code in request', () => {
+      let tubes = new Response(Data.SampleExtractionTubesWithSample).deserialize.assets
+      tubes[0].cost_code = null
+      let json = Actions.sampleExtractionTubeJson(tubes)
+      let tube = json[0]
+      expect(tube.cost_code).not.toBeDefined()
```
stevieing, line 136: I think cost code should be part of the request not the tube?? so if you do 
...
```
const [{ request }] = Actions.sampleExtractionTubeJson(tubes)
expect(request.cost_code).toBeNull()
```
...

stevieing (changes requested):
Hopefully I haven't misunderstood. Added some simplifications and a fix for the test.

tests/unit/store/traction/pacbio/requests/actions.spec.js
BenTopping: yes forgot to update this after resolving merge conflicts

src/store/traction/pacbio/requests/actions.js
BenTopping: The spread operator for assigning the cost code is needed because we dont want to include the cost code into the request if it is null /empty. This is because the backend will not default null values it only defaults if the value doesnt exist. So in this case it checks if the cost code is null and only includes it into the request if it is not null

stevieing: OK. But we can still simplify it a bit. Like so:

```
return tubes.map(
    ({
      library_type,
      estimate_of_gb_required,
      number_of_smrt_cells,
      barcode,
      study_uuid: external_study_id,
      sample_uuid: external_id,
      cost_code,
      fields: { sanger_sample_id: name, sample_common_name: species },
      ...t
    }) => ({
      sample: { name, species, external_id },
      request: {
        external_study_id,
        library_type,
        estimate_of_gb_required,
        number_of_smrt_cells,
        ...(cost_code ? {cost_code} : null)
      },
      tube: { barcode },
    }),
  )

```

BenTopping: 👍

stevieing (approved):
Just failing on Lint so I think you can remove the reference to t in actions.

PR #627, opened 2021-07-01
[WIP] Gpl 878 1 plate selection for pool

BenTopping:
Closes #

Notes TODO:
- cypress tests 
- mutation unit tests
- getter / view for list of requests in pacbioLabwareSelectedList
- add test to jsonApi for the relationships exist check (change in jsonApi is because we want to include relationships in wells but not all wells have relationships, so we want the includeRelationships flags true but it breaks without the conditional i added)

(probably also need to merge in main feature branch and pretty the code)


stevieing:
List of things to do:
fix cypress tests
getter / view for list of requests in pacbioLabwareSelectedList
modify PacbioSelectedWellItem to PacbioWellItem and isolate it (wellId to Id, import requests, request computed property)
modify PacbioSelectedPlateItem to PacbioPlateItem and modify (plateId to Id, import wells, find well by location, merge well and plate map attributes before creating well component)
modify PacbioLabwareSelectedList to get plates and pass down props rather than id)

PR #629, opened 2021-07-08
Create request factory

stevieing:
Closes #

This is a pull request that has come out of my JavaScript learning. It will:
- remove the need for a Request Vue component
- create a request factory which is now a POJO

TODO:
- add methods for updates
- add methods for deletes
- replace request in the ApiBuilder
- remove default includes and filters
- add documentation


stevieing:
Ready but not worth putting into develop until GPL_878 has been merged.

src/api/createRequest.js
```diff
@@ -0,0 +1,148 @@
+import axios from 'axios'
```
JamesGlover, line 1: I've had a reasonable amount of success dropping axios entirely, and just using the native `fetch` calls. The main differences I noted between that approach and axios is:
1) Only actual network errors throw, so you'll need to check the status of the response yourself.
2) I needed to URI encode the parameters myself, which made me a bit sad.

Examples from where I did it:
```ruby
const plateRequest = async (plate_barcode) => {
  try {
    return await window.fetch(PLATE_PATH + encodeURIComponent(plate_barcode), {});
  } catch (error) {
    console.error(error);
    throw `Network Error: ${error.message}`;
  }
}
```

Then actually making that request:
```ruby
 // We've not loaded the plate yet, so go fetch it.
      try {
        const response = await plateRequest(plate_barcode);

        if (response.status === 200) {
          const plate = await extractPlateJson(response);
          commit("updatePlate", { ...plate, scanned: true });
          commit("updateScanPlate", { barcode: plate_barcode, id: plate.id });
          if (!plate.control) {
            dispatch("fetchBatches", { ids: plate.batches });
          }
        } else {
          const error_message = await extractErrors(response);
          throw error_message;
        }
      } catch (error) {
        // We catch and log most expected errors where they occur, and re-throw with
        // a user-friendly error message. This handles displaying that
        commit("updateScanPlate", { barcode: plate_barcode, errorMessage: error });
      }
```

Not suggesting we drop axios here, but this always seems to be the approach people recommend whenever anyone runs into issues with axios.

src/api/createRequest.js
```diff
@@ -0,0 +1,148 @@
...
+    return value && typeof value === 'object' && value.constructor === Object
+  }
+
+  /*
+   * @param {Any} value
+   * @returns boolean
+   */
+  const isString = (value) => {
+    return (typeof value === 'string' || value instanceof String) && value.length > 0
+  }
```
JamesGlover, line 34: I think both of these can be pulled out of the constructor, and just be defined in the file, otherwise you're creating a new function for each request.

src/api/createRequest.js
```diff
@@ -0,0 +1,148 @@
...
+
+        // if it is a string turn it into a parameter e.g. 'a.b.c' will become 'key=a.b.c'
+        // if parameter is defined it will be 'parameter[key]=value'
+        if (isString(value)) {
+          return parameter ? `${parameter}[${key}]=${value}` : `${key}=${value}`
+        }
+      })
+      .filter(Boolean)
+      .join('&')
+  }
```
JamesGlover, line 57: Think this one could be pulled out as well...

src/api/createRequest.js
```diff
@@ -0,0 +1,148 @@
...
+   * @return String
+   * Turns a list of parameters into a query string
+   */
+  const buildQuery = ({ filter = {}, include = '', fields = {} } = {}) => {
+    const queryString = parametersToString({ filter, include, fields })
+
+    // if the query string has any length then put a ? in front
+    // otherwise just return an empty string
+    return queryString.length > 0 ? `?${queryString}` : ''
+  }
```
JamesGlover, line 72: And this one

src/api/createRequest.js
```diff
@@ -0,0 +1,148 @@
...
+        // we also pass the key as the parameter
+        if (isObject(value)) {
+          return parametersToString(value, key)
+        }
+
+        // if it is a string turn it into a parameter e.g. 'a.b.c' will become 'key=a.b.c'
+        // if parameter is defined it will be 'parameter[key]=value'
+        if (isString(value)) {
+          return parameter ? `${parameter}[${key}]=${value}` : `${key}=${value}`
+        }
```
JamesGlover, line 53: If we pass in non-string parameters here, such as a number/bool (which would be possible for a filter) then the parameter will be silently ignored. Should we throw if we match neither of these conditions to make it clear what's happened? (Part of me wonders if you just lose the string check entirely an let the string encoder do what it will.)

JamesGlover (approved):
Overall looks good. Only suggestion is removing a few of the utility functions out of the request constructor.

src/api/Config.old.json
```diff
@@ -0,0 +1,121 @@
+[
```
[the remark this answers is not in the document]
stevieing: I just left it in there for comparision. I can remove it for now.

src/api/createRequest.js
stevieing: Happy to leave it for now but as everything is in one place it would be easy to update if we ever run into problems.

src/api/createRequest.js
stevieing: Yes. I did think about this at the start but wanted to see where we ended up first.

src/api/createRequest.js
stevieing: I think I wanted to limit what could be passed and we could review later. Think I might just add a throw for now so at least it is not failing silently.

PR #650, opened 2021-08-04
Bug fix for pacbio run pools list

BenTopping:
Pools list would not load if a library within one of the pools did not have a tag

Closes #

*
*
* ...


src/store/traction/pacbio/pools/getters.js
```diff
@@ -5,7 +5,7 @@ const getters = {
       const libraries = libraryIds.map((libraryId) => {
         const { id, type, request, tag } = state.libraries[libraryId]
         const { sample_name } = state.requests[request]
-        const { group_id } = state.tags[tag]
+        const { group_id } = state.tags[tag] || ''
```
JamesGlover, line 8: It might make more sense to use an empty object here, rather than a string, as we proceed to de-structure it.  Doesn't make a huge difference, just makes it easier to see at a glance that group_id will be undefined, not ''

stevieing (approved):

PR #664, opened 2021-08-13
Gpl 878 27 e2e tests for run and pool edit

### LINKED ISSUE #661: GPL-878-27 [BUG] - As developers we need to add some e2e tests to ensure that any future changes do not introduce more bugs
**Describe the bug**
As developers we need to add some e2e tests to ensure that any future changes do not introduce more bugs. Part of the recent work introduced new bugs which we were unable to detect before UAT as there was not enough test coverage. From previous experience these kind of bugs are best exposed by e2e tests. So we need to add tests to the following pages at a minimum:
- pool edit
- run edit

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

stevieing:
Closes #661

*
*
* ...


stevieing:
This has already picked up one bug for pool update which is successful whether there is an error or not.

tests/e2e/specs/pacbio_run_create.js
```diff
@@ -37,53 +68,48 @@ describe('Pacbio Run Create view', () => {
       .select('Do Not Generate')
       .get('#bindingKitBoxBarcode')
       .type('12345')
-      .get('button')
-      .contains('Update')
```
JamesGlover, line 41: Generally with e2e tests its better to use things like content, as it mimics the way the user looks at the page. It can get a bit tricky if you have multiple buttons with the same label, but then at least it forces you to work out how a user may distinguish between them.

tests/e2e/specs/pacbio_run_create.js
```diff
@@ -37,53 +68,48 @@ describe('Pacbio Run Create view', () => {
...
+    const dataTransfer = new DataTransfer()
+
+    cy.visit('#/pacbio/runs')
+    cy.get('button')
+      .contains('New Run')
+      .click()
+    cy.get('#sequencing-kit-box-barcode').type('Lxxxxx101826100123199')
+    cy.get('#dna-control-complex-box-barcode').type('Lxxxxx101717600123199')
+    cy.get('#system-name').select('Sequel IIe')
+    cy.get('.list-group-item')
```
JamesGlover, line 100: Similarly here I'd suggest fetching based on label, but it appears that Cypress doesn't actually contain such functionality out the box. Not worth addressing now.

tests/e2e/specs/pacbio_pool_edit.js
```diff
@@ -45,13 +45,17 @@ describe('Pacbio Pool Edit', () => {
...
-    // and I click on edit pool
-    // I should see the pool barcode
-    // and the plate
-    // and the correct tag set
-    // and I remove the insert size
-    // and I click on update pool
-    // I should see pool not valid
+    cy.visit('#/pacbio/pools')
+    cy.get('.pool')
+      .get('[data-action=edit-pool]')
```
JamesGlover, line 50: This get isn't scoped to '.pool' as get is always global, you want `find`

tests/e2e/specs/pacbio_run_edit.js
```diff
@@ -1,22 +1,77 @@
...
+      statusCode: 200,
+      body: { data: {} },
+    })
+
+    cy.visit('#/pacbio/runs')
+    cy.get('#editRun-7').click()
+    cy.get('ellipse')
+      .first()
+      .click()
+      .get('#movieTime')
```
JamesGlover, line 34: There's lots of examples of this scattered throughout the tests, but given what I mentioned about get always being global, I'm wondering if its a mistake to chain the methods like this, and if we should always go back to cy when dealing with a new element. It obviously doesn't make a practical difference, but maybe it reads clearer? Possibly one to get @stevieing thoughts on as well, as he seemed to establish the chaining pattern, so perhaps had good reason.

JamesGlover (changes requested):
Only important one to change is the get in place of find, as this was what lead to flaky tests before.

JamesGlover:
There also seems to be a failing test, assume this is related to the bug-fix?

JamesGlover (approved):

PR #667, opened 2021-08-31
Bug printing failure

stevieing:
Closes #

*
*
* ...

### DIFF
--- a/src/store/traction/pacbio/libraries/actions.js
+++ b/src/store/traction/pacbio/libraries/actions.js
@@ -4,16 +4,18 @@ const createLibraryInTraction = async ({ rootGetters, getters }, library) => {
   let tag_id = rootGetters['traction/tractionTags'].find((l) => l.group_id == library.tag.group_id)
     .id
   let body = {
-    type: 'library',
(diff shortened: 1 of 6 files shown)


src/store/traction/pacbio/libraries/actions.js
```diff
@@ -4,23 +4,21 @@ const createLibraryInTraction = async ({ rootGetters, getters }, library) => {
...
-        volume: library.volume,
-        concentration: library.concentration,
-        template_prep_kit_box_barcode: library.templatePrepKitBoxBarcode,
-        insert_size: library.insertSize,
-      },
-      relationships: {
-        request: { data: { type: 'requests', id: library.sample.id } },
-        tag: { data: { type: 'tags', id: tag_id } },
-      },
+    type: 'library',
```
JamesGlover, line 7: I think this is what @BenTopping was already discussing on slack, but I think you still need the data here, as the update request processing expects that payload to be provided under the data argument, but there should actually be a data attribute in the payload itself.

BenTopping (approved):

PR #671, opened 2021-09-06
Bug library creation from sample

BenTopping:
Closes #

* Update library from library page is now working
* Creating a library from samples page now creates a single plexed pool via the pool endpoint instead of going through the library end point

Note: 
Went down a bit of a rabbit hole with this one, but had to call it as this work is not priority.

We need to think about how we deal with single plexed pools:
- if you delete a library from a single plexed pool on the libraries page it will delete the library within the pool and not the pool itself. You then end up with empty pools in the pools tab. You might be able to solve this by adding a conditional to library deletion where if library is only library in pool then delete pool aswell. 
- Similar thing to above with updating a library, if you update a library from a single plexed pool you only update the concentration, volume, insert size and temp...barcode for the library not the pool. This can result in incorrect values down the line.

### DIFF
--- a/src/components/pacbio/PacbioLibraryCreate.vue
+++ b/src/components/pacbio/PacbioLibraryCreate.vue
@@ -70,7 +70,7 @@
         >
           <b-form-input
             id="library-templatePrepKitBoxBarcode"
-            v-model="library.templatePrepKitBoxBarcode"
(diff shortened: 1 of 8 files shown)


stevieing (approved):
All good. Haven't bothered checking it out locally. Good use of the new response features.

PR #683, opened 2021-11-22
DPL-178 Add the ability to create single pacbio libraries without tags

### LINKED ISSUE #682: DPL-178 - As a long read user I would like to be able to create a library without tags so that I can put samples through the ultra low read pipeline (C=S, V=5)
**User story**
As a long read user I would like to be able to create a library without tags so that I can put samples through the ultra low read pipeline

Craig / Steve / Iraad

**Acceptance criteria**
- [ ] When I create a library in the API if it has a single sample with no tags it should be successful
- [ ] When I select a sample and click on create library the tag drop down should have a blank option
- [ ] When I create a library from a sample and I have not added a tag the library should be created successfully

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #682 

* Updated PacbioLibraryCreateModal to have a empty value
* Updated CreateLibraryInTraction pacbio action to allow empty tag_id
* ...

### DIFF
--- a/src/store/traction/pacbio/libraries/actions.js
+++ b/src/store/traction/pacbio/libraries/actions.js
@@ -4,8 +4,8 @@ import { handleResponse } from '@/api/ResponseHelper'
 
 const createLibraryInTraction = async ({ rootState, rootGetters }, library) => {
   // Some duplication of code from createPool but this is for single library pool
-  let tag = rootGetters['traction/tractionTags'].find((l) => l.group_id == library.tag.group_id)
-  let tag_id = tag ? tag.id : ''
+  const tag = rootGetters['traction/tractionTags'].find((l) => l.group_id == library.tag.group_id)
+  const tag_id = tag ? tag.id : ''
 
   let body = {
     data: {
--- a/src/store/traction/pacbio/libraries/actions.js
+++ b/src/store/traction/pacbio/libraries/actions.js
@@ -4,8 +4,8 @@ import { handleResponse } from '@/api/ResponseHelper'
 
 const createLibraryInTraction = async ({ rootState, rootGetters }, library) => {
   // Some duplication of code from createPool but this is for single library pool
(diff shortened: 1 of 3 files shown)


src/store/traction/pacbio/libraries/actions.js
```diff
@@ -4,8 +4,8 @@ import { handleResponse } from '@/api/ResponseHelper'
 
 const createLibraryInTraction = async ({ rootState, rootGetters }, library) => {
   // Some duplication of code from createPool but this is for single library pool
-  let tag_id = rootGetters['traction/tractionTags'].find((l) => l.group_id == library.tag.group_id)
-    .id
+  let tag = rootGetters['traction/tractionTags'].find((l) => l.group_id == library.tag.group_id)
+  let tag_id = tag ? tag.id : ''
```
stevieing, line 8: you could use const here unless you are changing it.

tests/unit/components/pacbio/PacbioLibraryCreate.spec.js
```diff
@@ -45,20 +45,21 @@ describe('PacbioLibraryCreate.vue', () => {
   })
 
   it('must have tagOptions data', () => {
-    expect(modal.tagOptions).toEqual([])
+    expect(modal.tagOptions).toEqual([{ value: '', text: 'No tag' }])
   })
 
   describe('#createLibrary', () => {
-    let payload
+    let payload, payload_no_tag
```
stevieing, line 52: As far as I can see payload_no_tag is only used in 1 test so you could declare it in the test. Also the preferred case type in JavaScript is camelCase.

stevieing (approved):
All good. Just a couple of syntactical suggestions.

PR #688, opened 2021-12-03
Issue675/dpl 137 as a long read user i need to be able to add tags automatically to all libraries in a pool  v 4

JamesGlover:
Closes #

*
*
* ...


stevieing (approved):
All looks good to me. I set it up locally to see how it works.

PR #690, opened 2021-12-10
Issue676/dpl 138 upload concentrations for pool creation

### LINKED ISSUE #676: DPL-138 - As a long read user I need to be able to upload concentrations for all of the libraries in a pool (C=M, V=3)
**User story**
As a long read user I need to be able to upload concentrations for all of the libraries in a pool. The concentrations are provided by the TOL team in a csv file. It is time consuming and error prone to add these manually

Steve / Craig

**Acceptance criteria**
- [ ] Ability to upload a csv file of concentrations for wells in a plate
- [ ] When the file is uploaded the concentrations are assigned to a pool

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

JamesGlover:
Closes #676


src/mixins/Helper.js
```diff
@@ -7,14 +7,6 @@ import * as consts from '@/consts/consts'
 export default {
```
stevieing, line 7: Not sure that this helper file is very helpful. Probably needs to go the same way as alerts!

stevieing (approved):
All good. Just a few comments on things that could be refactored later. Definitely much better having the alert globablly.

stevieing (dismissed):
All good. Added a few comments about things that could be refactored later. Having the alert centrally is much better.
I updated the prettier ignore to ignore csvs.

src/mixins/Helper.js
JamesGlover: Agreed, almost went ahead and killed it completely, more than happy to do so.

stevieing: Definitely. Kill it off.

src/lib/csv/pacbio.js
```diff
@@ -0,0 +1,78 @@
+import { parse } from 'csv-parse/lib/sync'
+
+const castFloat = (value) => Number(value)
+const castString = (value) => String(value)
+
+// Handles the casting of columns. By default, unless specified here, a column
+// will be cast to a String
+const columnCasting = {
```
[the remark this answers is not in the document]
JamesGlover: It could do, but it's pretty closely tied to the underlying data models that I'm not sure there's much to gain. (Ie. A change in configuration would always be accompanied by changes in the code-base itself)

Maybe the casting is happening in the wrong place, and we should be performing it in the mutations? I had originally planned to cast the source column here as well, but ultimately decided that made more sense elsewhere (as it made error handling substantially easier) so maybe the same is true here.

PR #722, opened 2022-03-08
Issue716/dpl 298 add ability to add tubes to a pacbio pool {c v 4}

### LINKED ISSUE #716: DPL-298 - Add ability to add tubes to a PacBio pool (C=?, V=4)
**User story**
While pacbio pools support tubes, the interface doesn't make it easy through the new interface. This makes multiplexed pool creation from tubes impossible.

It should be possible to select tubes in a similar manner to selecting plates.

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**
- [ ] add a list of acceptance criteria here
- [ ] ...

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

JamesGlover:
Closes #716

Some associated traction-service changes

* Tubes can be found in the pool create interface
* Autotagging supports tubes
* CSV upload supports tubes
* Interface better handles resizing
* Only show tag list when requested

### DIFF
--- a/src/components/pacbio/PacbioLabwareFind.vue
+++ b/src/components/pacbio/PacbioLabwareFind.vue
@@ -1,7 +1,7 @@
 <template>
   <b-col>
     <b-form @submit.prevent="handleSubmit()">
-      <h3>Find plates</h3>
+      <h3>Find Labware</h3>
       <b-form-input
         v-model="enteredLabware"
         data-input="labware-find"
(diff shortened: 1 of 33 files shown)


stevieing (changes requested):
Still looking at the code but the main issue I am seeing is that when I try to import a tube it errors with `plates should be present, [object Object]`.Looking at the requests it seems to be trying to create plates and requests? Plates fails with a 422 and requests fail with a 500. The samples are created but not the tubes.

Maybe it would also be good to strip the whitespace as I noticed when I copied in a barcode with a space at the end it failed.

This was probably part of the original work for importing tubes but I didn't pick it up.

I am also thinking that we will probably need a view so people can see the tubes?

stevieing (changes requested):
Unfortunately I am still getting an error but it is different to the last one. It is 'plates should be present and an array'.
The tubes are also still being imported twice but not sure whether that is what is supposed to happen?

stevieing (approved):

PR #742, opened 2022-03-28
Tailwind integration

Sangeetha-Bheeman:
Closes #

*
*
* ...


package.json
```diff
@@ -3,16 +3,19 @@
...
+    "start": "npm-run-all -p serve serve:css",
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "test:unit": "vue-cli-service test:unit",
     "test:e2e": "vue-cli-service test:e2e",
     "lint": "vue-cli-service lint --max-warnings 3",
     "pretty": "prettier --check '**/*'",
-    "test": "vue-cli-service test:unit && vue-cli-service test:e2e"
+    "test": "vue-cli-service test:unit && vue-cli-service test:e2e",
+    "serve:css": "postcss src/styles/tailwind.css -o src/styles/index.css"
```
KatyTaylor, line 14: If you look at the Readme, most of these commands are described in there, in various sections. Do you think you could add something to the Readme to describe what `yarn run start` and `yarn run serve:css` will do? Thanks

package.json
```diff
@@ -23,6 +26,8 @@
     "https-proxy-agent": "^2.2.2",
     "jest-transform-graphql": "^2.1.0",
     "lodash-es": "^4.17.21",
+    "postcss": "^7",
+    "tailwindcss": "npm:@tailwindcss/postcss7-compat",
```
KatyTaylor, line 30: Just wondering why the format of this is different to all the other dependencies?

src/main.js
```diff
@@ -9,6 +9,7 @@ import store from './store'
 import BootstrapVue from 'bootstrap-vue'
 import { apolloProvider } from './graphql/client'
 import globalAlert from '@/mixins/globalAlert'
+import './styles/index.css'
```
KatyTaylor, line 12: Looks like this is a huge file - does this contain the tailwind framework? Is it correct to store it or should it get generated when we build the package?

tailwind.config.js
```diff
@@ -0,0 +1,925 @@
+module.exports = {
+  content: ['./src/**/*.{html,vue}'],
+  // Used in the Labware component for slots, but colour is built dynamically so we need to make sure it's not purged
```
KatyTaylor, line 3: I can't find the Labware component - is this your comment or copied over from STAN? Thanks

src/App.vue
```diff
@@ -1,35 +1,44 @@
 <template>
-  <b-container id="app" fluid>
-    <br />
+  <div class="flex flex-col min-h-screen">
     <!-- TODO: move this into a header component -->
```
KatyTaylor, line 3: I see you've created a 'Header' component. Do you still want this 'TODO' comment there? Or are you happy with what's in and out of the header component?

package.json
seenanair: There is an earlier dependency in one of the existing packages in Traction that relies on an older version of PostCSS. So we need to use postcss7-compat along with tailwind, that is intended to use for this purpose.

src/main.js
seenanair: Yes, you are right. The development build is usually large by design. According to tailwind documentation, to make the development environment more flexible, tailwind generates thousands of utility classes which most of which we probably don't use. But while buiding in production enviornment, Tailwind optimizes final build size by removing unused styles.

tailwind.config.js
seenanair: That's something that needs to be removed. will do it.

src/App.vue
seenanair: Removed

package.json
seenanair: Done

KatyTaylor (dismissed):
Hi @seenanair , thanks for answering all my comments. I've approved it now, although:
a) I can see you've got some merge conflicts now, and 
b) Steve said in stand up today that he was planning to have a look

PR #754, opened 2022-04-12
Added Loading target for automatic sample sheet generation

Sangeetha-Bheeman:
Closes #

*
*
* ...

### DIFF
--- a/src/components/pacbio/PacbioRunWellEdit.vue
+++ b/src/components/pacbio/PacbioRunWellEdit.vue
@@ -90,12 +90,15 @@
           ref="loadingTarget"
           v-model="currentWell.loading_target"
           placeholder="Loading Target (P1+ P2)"
+          type="number"
+          :step="0.05"
+          :formatter="formatLoadingTargetValue"
           @input="updateAdaptiveLoading"
         >
         </b-form-input>
       </b-form-group>
 
-    <b-form-group
+      <b-form-group
         id="useAdaptiveLoading-group"
         label="Use Adaptive Loading"
         label-for="useAdaptiveLoading"
@@ -104,11 +107,11 @@
           id="useAdaptiveLoading"
           ref="useAdaptiveLoading"
           v-model="currentWell.use_adaptive_loading"
-          placeholder="False" readonly
+          placeholder="False"
+          readonly
         >
         </b-form-input>
       </b-form-group>
-
     </b-form>
 
     <b-table id="wellPools" stacked :items="currentWell.pools" :fields="wellPoolsFields">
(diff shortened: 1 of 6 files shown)


src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -79,6 +79,39 @@
         >
         </b-form-input>
       </b-form-group>
+
+      <b-form-group
+        id="loadingTarget-group"
+        label="Loading Target (P1+ P2): (0 to 1) "
```
sdjmchattie, line 85: Is it intentional to have a space on the right of the `+` but not on the left?

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -79,6 +79,39 @@
...
+        label-for="loadingTarget"
+      >
+        <b-form-input
+          id="loadingTarget"
+          ref="loadingTarget"
+          v-model="currentWell.loading_target"
+          placeholder="Loading Target (P1+ P2)"
```
sdjmchattie, line 92: Same here if you planned to change the spacing around the `+`.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -161,6 +194,8 @@ export default {
         ],
       },
       ccsAnalysisOutputOptions: ['Yes', 'No'],
+      regexToMatch: /(\d{0}\.\d{1,2})$/,
```
sdjmchattie, line 197: Not sure I understand the `\d{0}`.  The zero means don't try to match any digits, which is the same as leaving that out of the regex entirely.

sdjmchattie: Also, do you really want to fail to match values that have more than 2 decimal digits?  You can see this regex in action here:  https://regex101.com/r/p8kbkf/2

Note how it matches only the decimal point and two digits.  It also matches the last section where more than one decimal place was used and it allows alphabet characters before the final decimal section.  So this may not do what you think it is doing.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -179,6 +214,23 @@ export default {
...
+      if (this.currentWell.loading_target) {
+        this.currentWell.use_adaptive_loading = 'True'
+      } else {
+        this.currentWell.use_adaptive_loading = 'False'
+      }
+    },
+    formatLoadingTargetValue(val) {
+      if (val) {
+        if (this.regexToMatch.test(val)) {
+          return val
```
sdjmchattie, line 227: See if you had `abc.45` in the val, that's what you'd get back from this format function because it matches the regex.  If you want help forming a better regex, let me know.  I've got a lot of experience with them.

sdjmchattie (changes requested):

sdjmchattie (approved):

PR #757, opened 2022-04-19
Develop into Master

stevieing:
Closes #

*
*
* ...


.github/workflows/asset_dispatch.yml
```diff
@@ -16,11 +16,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-
+      - name: Export node version
+        id: node_version
+        run: echo ::set-output name=NODE_VERSION::$(cat .nvmrc)
```
sdjmchattie, line 21: Is this new syntax?  It's good!  But we had to use a workaround with environment variables in the past and still do in some workflows: https://github.com/sanger/crawler/blob/develop/.github/workflows/automated_release.yml#L61-L63

sdjmchattie, line 21: Is this new syntax?  It's good!  But we had to use a workaround with environment variables in the past and still do in some workflows: https://github.com/sanger/crawler/blob/develop/.github/workflows/automated_release.yml#L61-L63

src/api/ResponseHelper.js
```diff
@@ -38,11 +38,11 @@ const parseErrorArray = (errors) =>
  * @param {Object} data e.g. { data: { id: 1}}
  * @returns { Boolean, {Object}, String} { success, data, errors } e.g. { success: true, data: {id: 1}} or {success: false, errors: 'there was an error'}
```
sdjmchattie, line 39: Looks like this might be out of date.

src/api/ResponseHelper.js
```diff
@@ -38,11 +38,11 @@ const parseErrorArray = (errors) =>
...
  */
-const newResponse = ({ success, data, error }) => ({
+const newResponse = ({ success, data, error, errors }) => ({
```
sdjmchattie, line 41: Why does this handle both a singular error and the plural errors?  I feel there's an inefficiency somewhere in that.

harrietc52 (approved):
Just scanned over from a high level - agree with Stuart's comment on errors/ error

.github/workflows/asset_dispatch.yml
stevieing: I keep being asked about this. I copied it from Ben so not sure the history but it works. Probably worth having a shared workflow.

src/api/ResponseHelper.js
stevieing: OK. I will add it as a note for changing.

src/api/ResponseHelper.js
stevieing: I suspect they are different kinds of errors so not sure it is an efficiency situation. More likely the errors need renaming. It might be that error is the error returned from axios e.g. network error and errors are probably returned from the service. They have a different structures. I will add a note to it to come back to later.

Issue #734, opened 2022-03-18
DPL-320 - As Long Read we need to be able to print multiple barcodes with suffixes to allw tubes to be barcoded off LIMS [C=M, V=4]


stevieing:
**User story**
As Long Read we need to be able to print multiple barcodes with suffixes to allw tubes to be barcoded off LIMS. 

**Who are the primary contacts for this story**
e.g. John S (don't include surnames in public repos)

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] add a list of acceptance criteria here
- [ ] ...

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

harrietc52:
Mock up's [here](https://lucid.app/lucidchart/6450b990-71b2-4dbc-ae7f-1f013b1cf6ff/edit?invitationId=inv_c9f94f33-48eb-47e2-93db-1310b836139b)

![Screenshot 2022-04-25 at 14 32 49](https://user-images.githubusercontent.com/11446617/165099483-80e43982-60de-425a-92cc-129198a4257a.png)

PR #766, opened 2022-04-27
DPL-320

harrietc52:
**Deploy label_template script on deployment**

Closes https://github.com/sanger/traction-ui/issues/734


src/config/PrinterList.json
```diff
@@ -1 +1,7 @@
-["g216bc", "h105bc"]
+[
+  { "printerName": "g216bc", "printerType": "toshiba" },
+  { "printerName": "h105bc", "printerType": "toshiba" },
+  { "printerName": "stub", "printerType": "squix" },
+  { "printerName": "aPrinterName", "printerType": "squix" },
+  { "printerName": "ssrtubebc-sq1", "printerType": "squix" }
+]
```
harrietc52, line 7: This is what I was explaining the other day about adding the printerType here. Other options are to the add the labelTemplateName here, but that is a variable that should be preferable stored in Deployment project config, and therefore a VUE_APP variable, which can't be access from the JSON config file. So options:
1. Leave it as above, where the `printerType` is fixed and won't change, so it is okay to be hard coded. Then based off the printer type, access the correct labelTemplateName in code, which can be stored in the deployment project as a variable. 
2. Replace `printerType` here with `labelTemplateName`, which would have to be hardcoded here. 
3. Something else that I can't think of

harrietc52: @stevieing @JamesGlover  ^

harrietc52: Let me know your thoughts. Happy to update to option 2

tests/e2e/specs/visit_label_printing_page.js
```diff
@@ -0,0 +1,35 @@
...
+    cy.contains('Print Labels')
+    cy.contains('List of barcodes to be printed:')
+    cy.contains('Number of labels:')
+    cy.contains('aBarcode-L1')
+    cy.contains('aBarcode-L2')
+    cy.contains('aBarcode-L3')
+    cy.contains('Printer: stub')
+    cy.contains('Copies: 2')
+
+    cy.contains('OK').should('not.be.disabled')
```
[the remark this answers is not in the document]
harrietc52: Or @andrewsparkes, this is the last thing todo for this PR

src/config/PrinterList.json
stevieing: Based on our discussion you no longer need this as print my barcode will handle this. You can just go back to the array of printers as we are currently using a single template. I suspect this will change if we start to use this for other purposes but we can cross that bridge when we come to it.

tests/e2e/specs/visit_label_printing_page.js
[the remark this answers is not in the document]
stevieing: You can use intercept. See https://github.com/sanger/traction-ui/blob/develop/tests/e2e/specs/pacbio/import_samples_from_samples_extraction.js

.env.development
```diff
@@ -1,5 +1,7 @@
 VUE_APP_TRACTION_BASE_URL=http://traction
 VUE_APP_TRACTION_GRAPHQL_URL=http://traction/graphql
-VUE_APP_PRINTMYBARCODE_BASE_URL=http://printmybarcode
+VUE_APP_PRINTMYBARCODE_BASE_URL=http://localhost:3000
```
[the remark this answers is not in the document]
andrewsparkes: so you'd have a an .env.development file and an .env.development.local file and the variables with localhost values go in the .local file and it picks up the content of both files when running in development (hierarchy)?
would you check the .local file into the repo still?

.env.development
```diff
@@ -3,3 +3,4 @@ VUE_APP_TRACTION_GRAPHQL_URL=http://traction/graphql
 VUE_APP_PRINTMYBARCODE_BASE_URL=http://printmybarcode
 VUE_APP_SAMPLEEXTRACTION_BASE_URL=http://sampleextraction
 VUE_APP_LOG=true
+VUE_APP_TUBE_LABEL_TEMPLATE_NAME=traction_tube_label_template_v2
```
andrewsparkes, line 6: Not related to this line but did we want the .local versions of the env files in the end? Checked into the repo I mean..

tests/unit/components/labelPrinting/LabelPrintingModal.spec.js
```diff
@@ -0,0 +1,87 @@
...
+    expect(wrapper.find('.modal')).toBeTruthy()
+  })
+
+  it('will have an button component', () => {
+    expect(wrapper.find('.btn').element).toBeTruthy()
+    expect(wrapper.find('.btn').text()).toEqual('Print Labels')
+  })
+
+  describe('modal', () => {
+    it('displays the barcodes to be printed name', () => {
```
andrewsparkes, line 32: displays the barcodes to be printed name =>
displays the barcodes to be printed

tests/unit/store/printMyBarcode/index.spec.js
```diff
@@ -0,0 +1,25 @@
...
+    expect(Store.getters).toBeDefined()
+  })
+  it('has mutations', () => {
+    expect(Store.mutations).toBeDefined()
+  })
+  it('has actions', () => {
+    expect(Store.actions).toBeDefined()
+  })
+  it('has saphyr modules', () => {
+    expect(Store.modules.saphyr).toBeDefined()
```
andrewsparkes, line 17: Why are we testing the Store has these things that we aren't directly using?
And not printers and suffixes that we are.

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
...
+  methods: {
+    setSuffixOptions() {
+      let suffixOptions = this.$store.getters.suffixes.map((obj, index) => ({
+        value: index + 1,
+        text: obj['one_character_name'] + ' (' + obj['potential_label_name'] + ')',
+        char: obj['one_character_name'],
+      }))
+      suffixOptions.unshift({ value: null, text: MESSAGE_SELECT })
+      this.suffixOptions = suffixOptions
+    },
```
JamesGlover, line 142: I think this can be greatly simplified:

- There isn't really any reason to go via the store here. The data is entirely static, so you can simply `import SuffixList from '@/config/SuffixList'` within the script tags and use it directly.
- I don't think you need to worry about setting an id, and looking up the value. You can just set the value of the option to be the character itself and use it directly. (Obviously rename the variables)
- Also, as this doesn't change, it may be better set as a prop, rather than as data.

JamesGlover (commented):
Only looked at some of the bit so far, but have to head off. Thought I'd release my comments so far.

src/components/labelPrinting/LabelPrintingForm.vue
JamesGlover: Tried to find the documentation of using a function to construct default values for a prop, but its not readily apparent on the props section of the vue docs. Instead it seems confined to a comment in the validation section:
https://vuejs.org/guide/components/props.html#prop-validation

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
...
+const MESSAGE_SELECT = 'Please select a option'
+
+export default {
+  name: 'LabelPrintingForm',
+  components: {
+    LabelPrintingModal,
+  },
+  data() {
+    return {
+      form: {
```
stevieing, line 106: Mabe if the move the creation of the empty form to a method which creates a new object e.g. 
```
newForm = { barcode: null,
selectedSuffixId: null,
selectedNumberOfLabels: null,
selectedPrinterId: null,
copies: null }
```

then when you do reset you could just recreate the object.

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
...
+      }
+      return listSuffixedBarcodes
+    },
+    printer() {
+      return this.printerOptions[this.form.selectedPrinterId]
+    },
+    onReset(event) {
+      event.preventDefault()
+      // Reset our form values
+      this.form.barcode = null
```
stevieing, line 181: From above you could then just do `this.form = newForm` or something similar. If that works?

stevieing (changes requested):
I have left a fair few comments. Hopefully they make sense?? You can generally use const unlesss you are planning to mutate.

src/store/printMyBarcode/actions.js
```diff
@@ -0,0 +1,75 @@
...
+    labelTemplateName,
+  )
+
+  const promise = request.create({ data: payload })
+
+  const response = await handleResponse(promise)
+
+  if (!response.success) {
+    if (response.data.errors && response.data.errors.length != 0) {
+      let errors = response.data.errors.map((e) => e.source.pointer + ' ' + e.detail).join(', ')
```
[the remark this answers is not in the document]
JamesGlover: Yeah, so reading the json-api spec:
[tubeLabelTemplateName](https://jsonapi.org/format/#error-objects)
It sounds like pointer is the attribute that was causing the problem, and detail:

> detail: a human-readable explanation specific to this occurrence of the problem. Like title, this field’s value can be localized.

Which is a little vague if it should fully describe the error in isolation. I'm assuming in the case of traction itself it usually does, whereas I'm assuming it isn't for PMB. (Although I'd be amazed if we had consitency)

src/store/printMyBarcode/actions.js
```diff
@@ -0,0 +1,75 @@
...
+      return {
+        success: false,
+        errors: errors,
```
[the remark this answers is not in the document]
JamesGlover: I thought that at first as well, but I think as things stand we're only augmenting the errors with the pointer in cases where its avaialble, otherwise we fall back to the default error handling of the responseHandler.

There's a smell here definitely, but I think it might mean us needing to fix PMB errors.

Tempted to suggest we simplify this here at the expense of slightly funky errors, especially as we shouldn't be sending invalid payloads to PMB anyway.

tests/unit/store/printMyBarcode/index.spec.js
```diff
@@ -0,0 +1,25 @@
+import Store from '@/store/traction/index'
```
JamesGlover, line 1: This isn't testing the store object that the filename implies.

JamesGlover (changes requested):

src/components/labelPrinting/LabelPrintingForm.vue
harrietc52: I think props get passed in from the parent component, so not sure props would be best fit? But I will simplify it and leave as data for now

harrietc52: I have got rid of the id, it just means there can't be a placeholder 'select a suffix', as this needs the value of null

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
...
+    },
+    setPrinterNames() {
+      let printerOptions = this.$store.getters.printersWithType.map((obj, index) => ({
+        value: index + 1,
+        text: obj.printerName,
+        type: obj.printerType,
+      }))
+      printerOptions.unshift({ value: null, text: MESSAGE_SELECT })
+      this.printerOptions = printerOptions
+    },
```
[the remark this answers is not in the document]
harrietc52: Have removed the type and id, but left in the store, as it was already in the store for other pipelines

src/components/labelPrinting/LabelPrintingModal.vue
```diff
@@ -0,0 +1,75 @@
...
+    <b-modal
+      id="labelPrintingModal"
+      ref="labelPrintingModal"
+      size="sm"
+      title="Print Labels"
+      :static="isStatic"
+      @ok="sendPrintRequest"
+    >
+      <h3>List of barcodes to be printed:</h3>
+      <hr />
```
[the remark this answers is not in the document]
harrietc52: Removed the hr for now

harrietc52 (commented):
- Add summary instead of model @andrewsparkes 
- Better handle responses @harrietc52

src/store/printMyBarcode/actions.js
```diff
@@ -0,0 +1,75 @@
...
+
+const getLabelTemplateName = (getters) => {
+  return getters.tubeLabelTemplateName
+}
+
+const actions = {
+  printJobV2,
+}
+
+export { printJobV2, createPrintJobJsonV2 }
```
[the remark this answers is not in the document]
harrietc52: Thats a good idea. We'll keep is as v2 for now, until the v1 requests are updated to also use v2. Then we'll rename the v2 exported values to not include v2

src/store/printMyBarcode/actions.js
```diff
@@ -0,0 +1,75 @@
...
+    params.copies,
+    labelTemplateName,
+  )
+
+  const promise = request.create({ data: payload })
+
+  const response = await handleResponse(promise)
+
+  if (!response.success) {
+    if (response.data.errors && response.data.errors.length != 0) {
```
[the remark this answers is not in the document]
harrietc52: I think that might only be available in Vue 3 unfortunately?

tests/unit/components/labelPrinting/LabelPrintingForm.spec.js
```diff
@@ -0,0 +1,199 @@
+import LabelPrintingForm from '@/components/labelPrinting/LabelPrintingForm'
```
stevieing, line 1: It feels like things are being tested twice here. You have a thorough e2e test which fills in all of the fields and then they are being checked again here.

stevieing, line 1: It feels like things are being tested twice here. You have a thorough e2e test which fills in all of the fields and then they are being checked again here.

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
+<template>
```
stevieing, line 1: It is more traditonal for css ids and classes to be kebab case. e.g. barcode-input-group

stevieing, line 1: It is more traditonal for css ids and classes to be kebab case. e.g. barcode-input-group

tests/e2e/specs/visit_label_printing_page.js
```diff
@@ -0,0 +1,72 @@
+// https://docs.cypress.io/api/introduction/api.html
+
+describe('Label Printing page', () => {
+  beforeEach(() => {
+    cy.visit('#/label-printing')
+
+    cy.get('#barcode_input').type('aBarcode')
```
stevieing, line 7: One of the cool things I learnt recently is the use of data attributes. Data attributes are a way to separate styling from data. https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes. I have used them in newer e2e tests and I have found them useful. https://github.com/sanger/traction-ui/blob/develop/tests/e2e/specs/pacbio/pacbio_pool_create.js

stevieing, line 7: One of the cool things I learnt recently is the use of data attributes. Data attributes are a way to separate styling from data. https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes. I have used them in newer e2e tests and I have found them useful. https://github.com/sanger/traction-ui/blob/develop/tests/e2e/specs/pacbio/pacbio_pool_create.js

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
...
+  <b-container>
```
stevieing, line 2: I was wondering whether it might be an idea to have defaults for the fields? We know that they need to be selected but then we validate them. If there were defaults there would be no need to validate.

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
...
+    setPrinterNames() {
+      let printerOptions = this.$store.getters.printers.map((name) => ({
+        text: name,
+      }))
+      this.printerOptions = printerOptions
+    },
+    suffixedBarcodes() {
+      var listSuffixedBarcodes = []
+
+      if (this.form.barcode && this.form.selectedSuffix && this.form.selectedNumberOfLabels <= 9) {
```
stevieing, line 153: There is a min and max on the field but we are also validating it? The validation is that it is <=9 but the max is 10.

tests/unit/components/labelPrinting/LabelPrintingForm.spec.js
```diff
@@ -0,0 +1,199 @@
...
+          form: { selectedPrinterName: 'stub' },
+        })
+
+        const evt = {
+          preventDefault: () => {
+            return {}
+          },
+        }
+        labelPrintingForm.onReset(evt)
+        expect(labelPrintingForm.form.barcode).toEqual(null)
```
stevieing, line 143: You could always have a method on form is empty ...

stevieing (approved):
Well done for completing this one. It ended up being quite complex but hopefully will be useful. Lots more comments. Purely personal taste and advisory. I am more than happy to enact some of them if you are moving on to other things.

tests/unit/components/labelPrinting/LabelPrintingForm.spec.js
JamesGlover: Are you just talking about the `wrapper.find('#...').exists()` tests here? I think I'd agree about those. But generally speaking I don't see harm in covering stuff via both unit tests and e2e as they are great at surfacing different things.

src/components/labelPrinting/LabelPrintingForm.vue
JamesGlover: Min max don't actually limit the inputs, and only get validated when you trigger the native HTML5 validation via triggering a form submit. All they do outside of this context is limit the range of the spinner.

src/components/labelPrinting/LabelPrintingForm.vue
harrietc52: Copies and number of labels could both be 1, but might be odd for `barcode` and `printer` fields to have defaults

src/components/labelPrinting/LabelPrintingForm.vue
harrietc52: Ye the min and max on b-form-input doesn't prevent somebody from typing in directly "100". The max of 10 is for the number of copies, so a different field which doesn't have other validation

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -0,0 +1,194 @@
...
+        for (let i = 0; i < this.form.selectedNumberOfLabels; i++) {
```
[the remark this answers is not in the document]
harrietc52: Sorry not sure I follow, it's currently creating a list where the index is used as the unique key  and not sure performance is an issue here with max 9 labels

tests/e2e/specs/visit_label_printing_page.js
harrietc52: Oh cool ye. Definitely useful for a complicated UI, maybe for this simple form not worth the effort

tests/unit/components/labelPrinting/LabelPrintingForm.spec.js
harrietc52: cool I will remove the `wrapper.find('#...').exists()` but leave the rest

JamesGlover (approved):

PR #770, opened 2022-05-05
Changes to make adaptive loading editable and show/hide the loading t…

Sangeetha-Bheeman:
…arget

Closes #

*
*
* ...

### DIFF
--- a/src/api/PacbioRun.js
+++ b/src/api/PacbioRun.js
@@ -139,6 +139,7 @@ const createWellsPayload = (wells, plateId) => {
       pre_extension_time: well.pre_extension_time,
       binding_kit_box_barcode: well.binding_kit_box_barcode,
       loading_target_p1_plus_p2: well.loading_target_p1_plus_p2,
+      use_adaptive_loading: well.use_adaptive_loading,
       relationships: {
         plate: {
           data: {
--- a/src/api/PacbioRun.js
+++ b/src/api/PacbioRun.js
@@ -179,7 +179,7 @@ const updateRunPayload = (run) => {
 }
 
 const updateWellPayload = (well) => {
-  let poolsAttributes = well.pools.map((l) => {
+  const poolsAttributes = well.pools.map((l) => {
     return { type: 'pools', id: l.id }
   })
 
@@ -188,15 +188,7 @@ const updateWellPayload = (well) => {
       id: well.id,
       type: 'wells',
       attributes: {
-        row: well.row,
-        column: well.column,
-        movie_time: well.movie_time,
-        on_plate_loading_concentration: well.on_plate_loading_concentration,
-        generate_hifi: well.generate_hifi,
-        ccs_analysis_output: well.ccs_analysis_output,
-        pre_extension_time: well.pre_extension_time,
-        binding_kit_box_barcode: well.binding_kit_box_barcode,
-        loading_target_p1_plus_p2: well.loading_target_p1_plus_p2,
+        ...well,
       },
       relationships: {
         pools: {
--- a/src/api/PacbioRun.js
+++ b/src/api/PacbioRun.js
@@ -139,7 +139,6 @@ const createWellsPayload = (wells, plateId) => {
       pre_extension_time: well.pre_extension_time,
       binding_kit_box_barcode: well.binding_kit_box_barcode,
       loading_target_p1_plus_p2: well.loading_target_p1_plus_p2,
-      use_adaptive_loading: well.use_adaptive_loading,
       relationships: {
         plate: {
           data: {
--- a/src/components/pacbio/PacbioRunWellEdit.vue
+++ b/src/components/pacbio/PacbioRunWellEdit.vue
@@ -149,7 +149,6 @@
 
 <script>
 import { mapMutations, mapGetters, mapActions } from 'vuex'
-import _ from 'lodash'
 
 export default {
   name: 'WellModal',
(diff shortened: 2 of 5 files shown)


src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -93,22 +109,6 @@
           type="number"
```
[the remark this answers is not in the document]
Sangeetha-Bheeman: I agree with removing the formatter, but the value will not be limited to two decimal places and hence retained the formatter by making it lazy.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -215,23 +240,25 @@ export default {
...
         } else {
-          this.loadingTargetValue = parseFloat(val / 100).toFixed(2)
           return isNaN(this.loadingTargetValue) ? 0 : this.loadingTargetValue
         }
       }
     },
+    disableAdaptiveLoadingInput() {
+      this.disableAdaptiveLoading = true
+      this.currentWell.loading_target_p1_plus_p2 = ''
+      this.use_adaptive_loading = 'False'
```
stevieing, line 255: Is this attribute still relevant? I thought disableAdaptiveLoading replaced use_adaptive_loading?

src/store/traction/pacbio/runs/actions.js
```diff
@@ -25,8 +26,9 @@ const buildWell = ({ state }, position) => {
   let generate_hifi = generateHiFiDefault(state.currentRun.system_name)
```
stevieing, line 26: You can use const rather than let here as they are not being mutated.

tests/unit/store/traction/pacbio/runs/actions.spec.js
```diff
@@ -197,8 +197,8 @@ describe('#buildWell', () => {
       binding_kit_box_barcode: '',
       pools: [],
       pre_extension_time: 2,
-      loading_target_p1_plus_p2: '',
-      use_adaptive_loading: 'False',
+      loading_target_p1_plus_p2: 0.85,
```
stevieing, line 200: Again here we are still sending use_adaptive_loading to the service. Not sure whether we need to?

stevieing (changes requested):

src/components/pacbio/PacbioRunWellEdit.vue
Sangeetha-Bheeman: I am using this to set the 'disabled' attribute of the input box to true/false

tests/unit/store/traction/pacbio/runs/actions.spec.js
Sangeetha-Bheeman: This is just setting the default attributes while creating/building a new well. It will not be sent back to the service.

src/components/pacbio/PacbioRunWellEdit.vue
stevieing: OK. You could follow James' comments above about creating a computed field then you would not need to set the value.

```
{ // in computed
  useAdaptiveLoading: (component) => component.currentWell.loading_target_p1_plus_p2 !== '' 
}
```

tests/unit/store/traction/pacbio/runs/actions.spec.js
Sangeetha-Bheeman: It is just assigning a default value when creating/building a new well. This value will not included in creating the request payload to the service.

stevieing: I understand. I think the point is that you do not need to it at all because it is a computed value based on the loading_target_p1_plus_p2 in the component. Unless I am missing something? More than happy to talk through it with you.

Sangeetha-Bheeman: Thank you for pointing it. Removed the lines.

src/components/pacbio/PacbioRunWellEdit.vue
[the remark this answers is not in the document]
stevieing: We have validation in the service? Do we need to validate in the front end?

src/api/PacbioRun.js
```diff
@@ -188,15 +188,7 @@ const updateWellPayload = (well) => {
...
-        row: well.row,
-        column: well.column,
-        movie_time: well.movie_time,
-        on_plate_loading_concentration: well.on_plate_loading_concentration,
-        generate_hifi: well.generate_hifi,
-        ccs_analysis_output: well.ccs_analysis_output,
-        pre_extension_time: well.pre_extension_time,
-        binding_kit_box_barcode: well.binding_kit_box_barcode,
-        loading_target_p1_plus_p2: well.loading_target_p1_plus_p2,
+        ...well,
```
JamesGlover, line 191: I love destructuring/spreading etc.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -247,15 +222,10 @@ export default {
       }
     },
     disableAdaptiveLoadingInput() {
-      this.disableAdaptiveLoading = true
       this.currentWell.loading_target_p1_plus_p2 = ''
       this.use_adaptive_loading = 'False'
       this.loadingTargetText = 'Adaptive loading disabled - Add loading target to enable'
```
JamesGlover, line 227: See above, I don't think this needs to be dynamic.

JamesGlover (changes requested):

stevieing (approved):

JamesGlover (approved):

PR #778, opened 2022-05-17
Develop into Master

stevieing:
Closes #

*
*
* ...


src/api/ResponseHelper.js
```diff
@@ -37,6 +37,7 @@ const parseErrorArray = (errors) =>
  * @param Boolean success
  * @param {Object} data e.g. { data: { id: 1}}
  * @returns { Boolean, {Object}, String} { success, data, errors } e.g. { success: true, data: {id: 1}} or {success: false, errors: 'there was an error'}
+ * TODO: some confusion on what the arguments are and what is returned. Need to review and update documentation.
```
harrietc52, line 40: Is this noted in a story anywhere to come back to?

src/components/Link.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <div @click="$router.push(`${link}`).catch((error) => routerError(error))">
+    <router-link :to="`${link}`" :class="getClass()">
+      {{ name }}
+    </router-link>
+  </div>
+</template>
+
+<script>
+export const ViewType = {
```
harrietc52, line 10: const want to be in uppercase?

src/components/Link.vue
```diff
@@ -0,0 +1,56 @@
...
+  Button: 1, //Display as dashboard button
+  MenuItem: 2, // Display as menu item
```
harrietc52, line 12: and camelCase?

tests/unit/views/Dashboard.spec.js
```diff
@@ -36,38 +36,38 @@ describe('Dashboard.vue', () => {
...
 
     it('will have a description', () => {
-      expect(box.find('.card-text').text()).toBeDefined()
+      expect(box.find('[data-attribute=description]').text()).toBeDefined()
     })
 
     describe('route buttons', () => {
       it('will have a reception button', () => {
-        let receptionButton = box.findAll('button').at(0)
+        let receptionButton = box.findAll('a').at(0)
```
harrietc52, line 52: How come findAll `a` ?

harrietc52 (approved):

src/api/ResponseHelper.js
stevieing: This was something raised by Stuart a while back with regard to passing errors and error at the same time. I think! So not too sure.

src/components/Link.vue
stevieing: This was something Seena added as part of the styling. I think it is like that because it is an enum.

tests/unit/views/Dashboard.spec.js
stevieing: We changed the buttons to hyperlinks and it broke the tests. It looks weird but I think this is the standard way in vue test utils to find stuff. I couldn't find a better way.

PR #779, opened 2022-05-19
Allow configuring default well metrics

Sangeetha-Bheeman:
Closes #

*
*
* ...


JamesGlover (changes requested):
Generally looks pretty solid.

1) The re-purposed run-name field isn't really applicable here. It does actually do something for the original run, so here just a plain header would be more suited.
2) As we've got a few defaults, the placeholder fields don't really work all that well anymore, and we might be better switching to explicit labels
3) Then things get a little more complicated, where I suggest some refactoring of defaults into `PacbioRun.js`. This should simplify a few other places, but happy to have a chat about this if my comments don't make sense.

I think most of these are a result of you continuing on with existing patterns you'd found. So this is mostly about how we can tidy things up, especially where some of the more recent changes have made the existing patterns/UX a little cumbersome.

src/store/traction/pacbio/runs/actions.js
```diff
@@ -1,36 +1,30 @@
...
+  state.currentRun.wellDefault.generate_hifi = PacbioRun.wellDefaults(
+    state.currentRun.system_name,
+  ).generate_hifi
+  const movie_time = state.currentRun.default_movie_time || state.currentRun.wellDefault.movie_time
+  const generate_hifi =
+    state.currentRun.default_generate_hifi || state.currentRun.wellDefault.generate_hifi
+  const ccs_analysis_output =
+    state.currentRun.default_ccs_analysis_output || state.currentRun.wellDefault.ccs_analysis_output
+  const pre_extension_time =
+    state.currentRun.default_pre_extension_time || state.currentRun.wellDefault.pre_extension_time
```
JamesGlover, line 21: I'm not quite sure why you have the fallback here, rather than just using the values under wellDefault. The advantage of that is this you can just spread wellDefault into the new object, rather than having to extract each value individually.

JamesGlover (changes requested):
The well default changes are nice, but there seem to be a few left-over bits of the old approach, which can be cleaned up, simplifying things significantly.

Also, I think there is some misunderstanding over some of the bits of the UI, but this may be easier to talk about over zoom.

JamesGlover (approved):

PR #798, opened 2022-06-20
Check release version

Sangeetha-Bheeman:
Closes #

*
*
* ...

### DIFF
--- a/.github/workflows/check_release_version.yml
+++ b/.github/workflows/check_release_version.yml
@@ -12,16 +12,23 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Get the latest release
-        id: last_release
-        uses: pozetroninc/github-action-get-latest-release@v0.5.0
+      - name: Get specific changed files
+        id: changed-files-specific
+        uses: tj-actions/changed-files@v23
         with:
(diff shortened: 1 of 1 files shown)


sdjmchattie (changes requested):
I like the use of the `changed-files-specific` action.  I wonder if the logic is a bit wrong so I've made some suggestions to try.

sdjmchattie (changes requested):
Looks good.  Nice and neat.  I wonder whether it would be worth creating a new branch for this and just putting one commit into it with the changes to the `check_release_version.yml` file only and skip all the testing commits.  This branch can then be deleted and the history will be cleaner in the git repo.

JamesGlover:
> Looks good. Nice and neat. I wonder whether it would be worth creating a new branch for this and just putting one commit into it with the changes to the `check_release_version.yml` file only and skip all the testing commits. This branch can then be deleted and the history will be cleaner in the git repo.

Github lets you do a squash-merge for that kind of thing, although I'm not sure if that lets you change the commit message, or if you'll end up with ones with ending in 'Testing', 'Testing', 'Testing'

Sangeetha-Bheeman:
Opened another PR, since this has many test commits.

PR #810, opened 2022-06-27
Develop into Master

stevieing:
Closes #

*
*
* ...


src/store/traction/pacbio/runs/mutations.js
```diff
@@ -6,14 +8,23 @@ const mutateRun = (key) => (state, val) => {
   state.currentRun[key] = val
 }
 
+const mutateWellDefaults = (key) => (state, val) => {
+  Vue.set(state.currentRun.wellDefaults, key, val)
```
harrietc52, line 12: What does `Vue.set` do here?

harrietc52 (approved):

src/store/traction/pacbio/runs/mutations.js
stevieing: > What does `Vue.set` do here?

Vue.set is a tool that adds a new property to an already reactive property and makes sure that property is also reactive. If you didn't use it the property would not show up in the ui. In Vue 3 it is no longer needed as you can make an object reactive and any new properties are reactive by default.

PR #834, opened 2022-07-07
Extract useful components

JamesGlover:
I've added some components, and performed a bit of refactoring as part of the receptions work.

I've extracted this out so @seenanair can use it, and will close and then rebase the rest of my PR.


src/components/shared/TractionMenuItem.vue
```diff
@@ -0,0 +1,24 @@
...
+      active ? 'text-white bg-sp' : 'text-sp',
+    ]"
+  >
+    <slot />
+  </li>
+</template>
+
+<script>
+export default {
+  name: 'TractionMenu',
```
seenanair, line 16: A doubt - Is it not 'TractionMenuItem'?

JamesGlover: You are correct. One of the flaws of using a previous component as a template.

JamesGlover: Fixed.

JamesGlover:
Documentation wise it might be worth having a consideration of storybook/Vue Styleguidist to make sure that whatever format we adopt works well with that. But yes, agreed.

PR #837, opened 2022-07-08
Issue706/dpl 278 ont reception import a plate or tube from sequencescape or samples extraction with a library type rebase

JamesGlover:
Closes #

*
*
* ...


README.md
```diff
@@ -1,4 +1,4 @@
-# Traction UI
+# ![logo](/public/traction-logo.svg) Traction UI
```
JamesGlover, line 1: Because why not

__mocks__/swrv.js
```diff
@@ -1,11 +1,122 @@
+const baseURL = import.meta.env.VITE_TRACTION_BASE_URL
```
JamesGlover, line 1: I think the Vue 2.7 upgrade might actually mean we can use mirage here instead. But my timebox ran out.

cypress.config.js
```diff
@@ -9,6 +9,6 @@ export default defineConfig({
     setupNodeEvents(on, config) {},
     specPattern: 'tests/e2e/specs/**/*.cy.{js,jsx,ts,tsx}',
     supportFile: 'tests/e2e/support/index.js',
-    baseUrl: 'http://localhost:3000/',
+    baseUrl: 'http://localhost:5174/',
```
JamesGlover, line 12: Vite 3 has changed the default port, so as not to clash with other frameworks (such as Rails)

index.html
```diff
@@ -21,6 +21,21 @@
...
+        text-align: center;
+        font-family: 'Helvetica Neue', Arial, system-ui, -apple-system, BlinkMacSystemFont,
+          'Segoe UI', Roboto, 'Noto Sans', sans-serif;
+        color: #718096;
+      }
+
+      #app-loading-pane-logo {
+        width: 100%;
+      }
+    </style>
```
JamesGlover, line 38: I've added the logo on the loading page. Kept the styles inline so everything loads quickly.

package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@sanger/ui-styling": "^4.0.0",
-    "@vue/composition-api": "^1.6.3",
+    "@vitejs/plugin-vue2": "^1.1.2",
```
JamesGlover, line 25: Part of the vue2.7 upgrade

package.json
```diff
@@ -55,8 +54,9 @@
     "npm-run-all": "^4.1.5",
     "postcss-cli": "^10.0.0",
     "prettier": "2.7.1",
+    "rollup-plugin-visualizer": "^5.7.0",
     "sass": "^1.53.0",
     "vitest": "^0.15.2",
-    "vue-template-compiler": "^2.6.12"
+    "vue-template-compiler": "^2.7.5"
```
JamesGlover, line 60: We should be able to drop this with the next release of vue-test-utils

src/components/PrinterModal.vue
```diff
@@ -3,7 +3,7 @@
     <traction-button
       id="printLabels"
       v-traction-modal.printerModal
-      variant="primary"
```
JamesGlover, line 6: Lots of this kind of thing. Variant didn't bother me too much, but I wanted the theme names to make sense in the context of a button. What is a successful button?

src/components/shared/DataTypeSelect.vue
```diff
@@ -0,0 +1,54 @@
+<template>
```
JamesGlover, line 1: I decided to part with the pattern in LibraryTypeSelect, and decided that the label should be managed independently. We should ideally adopt the same pattern in LibraryTypeSelect.

This separates out the field from its wider context, increasing reusability. (For example I can use it in auto-generated forms, or decide I need to put the label alongside the field, rather than above it)

I noticed that the libraryType select had already been picking up lots of configuration for this kind of thing.

src/components/shared/DataTypeSelect.vue
```diff
@@ -0,0 +1,54 @@
...
+  <traction-select
+    id="data-type"
+    :value="value"
+    :options="dataTypes"
+    @input="handleInput"
+  ></traction-select>
+</template>
+
+<script>
+import useSWRV from 'swrv'
```
JamesGlover, line 11: Ideally I'd like to pull a lot of the swrv contents out into a separate file, or, if not possible a mixin.

src/components/shared/LibraryTypeSelect.vue
```diff
@@ -1,42 +1,37 @@
...
-      <traction-select
-        id="library-type"
-        :class="getClass()"
-        :value="libraryType"
-        :options="libraryTypes"
-        @input="handleInput"
-      ></traction-select>
-    </traction-form-group>
-  </div>
+  <traction-field-group label="Library Type" label-for="library-type">
```
JamesGlover, line 2: This change means the field looks slightly out of place in:
1) PacbioSamplesExtractionReception - which will be removed in short order
2) The request metadata edit page - as the style of the label doesn't quite match up. However this is subtle, and I decided to leave the updating of that component to a later story.

JamesGlover: In its other contexts the updated style slots in just fine.

src/components/shared/LibraryTypeSelect.vue
```diff
@@ -52,19 +47,18 @@ export default {
...
-      type: Number,
-      default: 2,
-    },
-    labelAlign: {
-      type: String,
-      default: 'left',
-    },
-    labelPosition: {
-      type: [Number, String],
-      default: () => POSITION.Left,
```
JamesGlover, line 65: I've removed a lot of the customisation here. Ultimately I think this should be like the DataTypes select, and be solely concerned with the Select element itself.

src/main.js
```diff
@@ -1,23 +1,20 @@
-import 'bootstrap/dist/css/bootstrap.css'
-import 'bootstrap-vue/dist/bootstrap-vue.css'
+import './styles/legacy_bootstrap.scss'
+import './styles/index.css'
 
 import Vue from 'vue'
 import Vuex from 'vuex'
 import App from './App.vue'
+import VueRouter from 'vue-router'
```
JamesGlover, line 7: This was being imported in router.js before, but that meant we were mutating the global vue when loading our test. This prevented us from mocking the router.

src/main.js
```diff
@@ -1,23 +1,20 @@
...
 import router from './router'
 import store from './store'
 import globalAlert from '@/mixins/globalAlert'
-// Vue 2.7 (and 3) includes the composition API, so can remove this shortly.
-import VueCompositionAPI from '@vue/composition-api'
-import './styles/index.css'
```
JamesGlover, line 12: For once, a comment promising that something will happen 'shortly' turns out to be correct!

src/services/Sequencescape.js
```diff
@@ -1,6 +1,13 @@
+// This module has been superseded by Receptions/Sequencescape and can be removed
```
JamesGlover, line 1: I'll see if I can prepare a post flag branch ASAP.

stevieing (approved):

PR #851, opened 2022-07-15
Test latest tailwind

JamesGlover:
I've:
- Sanity checked the styles, I can't promise there are no issues, but nothing that jumped out
- I've fixed the warning

### DIFF
--- a/tailwind.config.cjs
+++ b/tailwind.config.cjs
@@ -1,10 +1,9 @@
-const defaultOptions = require('../traction-ui/node_modules/@sanger/ui-styling/tailwind.config')
+const defaultOptions = require('@sanger/ui-styling/tailwind.config')
+
 module.exports = {
   ...defaultOptions,
-  purge: {
-    ...defaultOptions.purge,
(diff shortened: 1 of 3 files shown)


tailwind.config.cjs
```diff
@@ -1,10 +1,9 @@
...
+const defaultOptions = require('@sanger/ui-styling/tailwind.config')
+
 module.exports = {
   ...defaultOptions,
-  purge: {
-    ...defaultOptions.purge,
-    content: ['./src/**/*.{html,vue}'],
-  },
+
+  content: [...defaultOptions.content, './src/**/*.{html,vue}'],
```
JamesGlover, line 6: Content is defined already, I'm assuming its passed through from tailwind itself.

seenanair: > Content is defined already, I'm assuming its passed through from tailwind itself.

 Looking good. 'purge' is not in use any more. In ui-styling package , content supports html and react(jsx,tsx) files. So, I think over riding  here for vue files is required.

seenanair (approved):

PR #858, opened 2022-07-21
Ensure we don't 'eat' network errors

JamesGlover:
This is very much a sticking plaster. A more extensive fix was resulting
in errors which were previously getting 'caught' instead failing
silently.

Rather than change the behaviour, I've instead ensured that the original
error propagates outwards. This won't actually fix the intermitted
failing cypress test, but should ensure we get a more useful error.

It also improves the network error messages users may see in reality.

Closes #

*
*
* ...

### DIFF
--- a/src/api/PromiseHelper.js
+++ b/src/api/PromiseHelper.js
@@ -13,9 +13,24 @@ const handlePromise = async (promise) => {
   try {
     rawResponse = await promise
   } catch (resp) {
-    let responseObject = resp.response
-    responseObject['data'] = responseObject['data']['data']
-    rawResponse = responseObject
+    if (resp.response) {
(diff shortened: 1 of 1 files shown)


stevieing (approved):
Good. It would obviously be better to use responseHelper but that would take a while to change and I haven't got time and you're leaving! Maybe it would be worth adding a comment to this to mark it as deprecated wrt to responseHelper?

JamesGlover:
> Good. It would obviously be better to use responseHelper but that would take a while to change and I haven't got time and you're leaving! Maybe it would be worth adding a comment to this to mark it as deprecated wrt to responseHelper?

There already is one at the top of the file

PR #859, opened 2022-07-21
Clean up flagged code

JamesGlover:
We have a large amount of code which is unused once
flags have flipped. This change removes it all.

Closes #

*
*
* ...


BenTopping:
Rebasing appears to break PRs when the difference is this big. Will close and rebuild

PR #865, opened 2022-07-26
Dpl 431 i need to be able to select the smrt link version

stevieing:
Closes #

*
*
* ...

### DIFF
--- a/src/api/PacbioRun.js
+++ b/src/api/PacbioRun.js
@@ -192,7 +192,7 @@ const createWellsPayload = (wells, plateId) => {
   }
 }
 
-// This is exactly the same as the create payload apart from the id
+// This is exactly the same as createRunPayload apart from the included id key here
 const updateRunPayload = (run) => {
   return {
     data: {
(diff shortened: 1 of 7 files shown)


tests/unit/api/PacbioRun.spec.js
```diff
@@ -3,6 +3,7 @@ import Response from '@/api/Response'
 import * as Run from '@/api/PacbioRun'
 import build from '@/api/ApiBuilder'
 import Api from '@/api'
+import { it } from 'vitest'
```
JamesGlover, line 6: Ahh, VScode keeps doing this, harmless, but ultimately unnecessary as these are available globally.  Still, makes me wonder if we should just import them explicitly the whole time.

JamesGlover (dismissed):

src/api/PacbioRun.js
```diff
@@ -189,6 +192,7 @@ const createWellsPayload = (wells, plateId) => {
   }
 }
 
+// This is exactly the same as the create payload apart from the id
```
sdjmchattie, line 195:
```suggestion
// This is exactly the same as createRunPayload apart from the included id key here
```

tests/unit/api/PacbioRun.spec.js
sdjmchattie: Generally explicit beats implicit and global, I would have thought.

sdjmchattie (dismissed):

sdjmchattie (approved):

PR #905, opened 2022-09-30
dpl- 496 traction input component

seenanair:
- New traction-input component using tailwind
- New traction-field-error component to display validation error message and icon 
- Removed bootstrap input coponent
- Replaced all bootstrap input usages with traction-input


src/components/shared/TractionFieldError.vue
```diff
@@ -0,0 +1,41 @@
...
+    dataAttribute: {
+      type: String,
+      default: '',
+    },
+    withIcon: {
+      type: Boolean,
+      default: false,
+    },
+  },
+}
```
stevieing, line 40: This looks good. Again might be worth a bit more documentation just to explain what is going on e.g. I am not really sure what the significance of withIcon is.

src/components/shared/TractionFieldGroup.vue
```diff
@@ -5,7 +5,13 @@
   - label: The label to apply
   - description: Longer text providing more help
   - component: the type of component to render (if not handled in the slot)
-  - componentProps: The configuration for the component
```
stevieing, line 8: This documentation is much better but I am still struggling to understand it. e.g. what is the traction field group for? You mention componentProps and what they are for but I am still slightly confused about what it means or maybe it is just me!

src/components/shared/TractionFieldGroup.vue
```diff
@@ -54,6 +62,9 @@ export default {
         this.$emit('input', value)
       }
     },
+    getAttribute() {
```
stevieing, line 65: could this be a computed property rather than a method? Not sure just asking the question.

src/components/shared/TractionInput.vue
```diff
@@ -0,0 +1,87 @@
+<template>
```
stevieing, line 1: This is great. It would be good to see a bit more documentation especially related to what each of the props actually does and how they are used.

src/components/shared/icons/TractionPassIcon.vue
```diff
@@ -0,0 +1,15 @@
+<template>
```
stevieing, line 1: The traction pass and fail icon look very similar and especially how they are called. Is there any way to combine them? Not sure that is a job for this story as it is not necessayr but maybe worth thinking about for the future?

stevieing (changes requested):
Really good piece of work. Just a few suggested changes mostly around documentation. I will also test this locally to see how it is working

seenanair:
@stevieing Thank you very much for your review comments. I will go through it. 👍

stevieing (commented):
I have tried this locally and everything seems to work correctly :)

seenanair:
> I have tried this locally and everything seems to work correctly :)
👍  :)
Thank you, steve.

src/components/pacbio/PacbioPoolLibraryEdit.vue
```diff
@@ -145,14 +148,18 @@ export default {
     ...mapMutations(['updateLibrary']),
     ...mapActions(['applyTags']),
     hasErrors(attribute) {
-      if (!this.library.errors) {
-        return null
+      if (!this.library.errors || !this.library.errors[attribute]) {
```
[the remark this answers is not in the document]
seenanair: Yes, there was no need to check for library.errors[attribute].length >0 as it is already included in truthy/falsy condition using !this.library.errors[attribute]

src/components/shared/TractionFieldGroup.vue
seenanair: I have added some more documentation. Please check if that is any good.

src/components/shared/TractionFieldGroup.vue
seenanair: Yes, you are right. Going through the documentation, it feels like getAttribute is better to be a computed property as it is just returning data from existing sources

src/components/shared/TractionInput.vue
seenanair: 👍

src/components/shared/icons/TractionPassIcon.vue
seenanair: Done. Please check any further changes required

src/components/pacbio/PacbioPoolLibraryEdit.vue
```diff
@@ -145,14 +145,21 @@ export default {
     ...mapMutations(['updateLibrary']),
     ...mapActions(['applyTags']),
     hasErrors(attribute) {
-      if (!this.library.errors) {
-        return null
+      //Check any errors exist in library for the given attribute
```
stevieing, line 148: This can be simplified to:
```
hasErrors(attribute) {
  return !this.library?.errors?.[attribute]
}
```
the ? is known as optional chaining which means you don't have to repeat stuff. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/components/pacbio/PacbioPoolLibraryEdit.vue
```diff
@@ -145,14 +145,21 @@ export default {
...
+      if (!this.library.errors || !this.library.errors[attribute]) {
+        return false
       }
-      return !this.library.errors[attribute]
+      return true
     },
     errorsFor(attribute) {
```
stevieing, line 154: It looks like errorsFor and hasErrors are practically the same but the reverse. Also here you can do 
`return this.library?.errors?.[attribute]`

src/components/pacbio/PacbioPoolLibraryEdit.vue
```diff
@@ -145,14 +145,21 @@ export default {
...
       return this.library.errors && this.library.errors[attribute]
     },
+    attributeValueExists(attribute) {
```
stevieing, line 157: again here you can do optional chaining.
`return this.library?.[attribute]?.length > 0`

stevieing (dismissed):
Approved. I have added a couple of things. personal choice. Don't affect behaviour.

src/components/pacbio/PacbioPoolLibraryEdit.vue
seenanair: Done

src/components/pacbio/PacbioPoolLibraryEdit.vue
seenanair: 'errorsFor' returns the error value itself  if any and 'hasErrors' gives a boolean check and return true/false depending on whether there is an error or not.
So, I have removed hasErrors which can in effect be replicated from errorsFor return value

stevieing (approved):

PR #910, opened 2022-10-05
Label printing changes

Sangeetha-Bheeman:
Closes #

*
*
* ...


src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -139,8 +141,9 @@ export default {
   methods: {
```
stevieing, line 141: A bit of documentation would be good to explain what is going on

stevieing (changes requested):
All seems ok. Needs a bit of documentation. I have also asked Ben to have a look at it.

stevieing (commented):
Also broken tests.

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -148,15 +151,25 @@ export default {
...
     },
     suffixedBarcodes() {
-      var listSuffixedBarcodes = []
+      let listSuffixedBarcodes = []
 
       if (this.form.barcode && this.form.selectedSuffix && this.form.selectedNumberOfLabels <= 9) {
-        for (let i = 0; i < this.form.selectedNumberOfLabels; i++) {
-          listSuffixedBarcodes.push(this.form.barcode.concat('-', this.form.selectedSuffix, i + 1))
+        let barcodes = this.form.barcode.split(/\r?\n|\r|\n/g)
+        //let suffix = this.form.selectedSuffix
```
BenTopping, line 158: is this suppose to be commented out?

tests/unit/components/labelPrinting/LabelPrintingForm.spec.js
```diff
@@ -43,7 +43,7 @@ describe('LabelPrintingForm.vue', () => {
 
   describe('methods', () => {
     it('setSuffixOptions ', async () => {
-      expect(labelPrintingForm.suffixOptions.length).toEqual(SuffixList.length)
+      expect(labelPrintingForm.suffixOptions.length - 1).toEqual(SuffixList.length)
```
BenTopping, line 46: Is this -1 because there is an additional option of 'No suffix'?

BenTopping: Might be better to compare the list values rather than just comparing the length as it will also make sure the values are correct

src/components/labelPrinting/LabelPrintingForm.vue
Sangeetha-Bheeman: Yes, I will remove the commented lines.

tests/unit/store/printMyBarcode/actions.spec.js
```diff
@@ -6,14 +6,23 @@ const formatDate = () => {
   return `${dd}-${mmm}-${yyyy.slice(2)}`
 }
 
+const getSuffix = (suffix, barcode) => {
+  return suffix ? barcode.slice(-5) : ''
```
BenTopping, line 10: is the suffix always the same length?

BenTopping: If not, might be better to get the text after the last hyphen if suffix always has a hyphen

Sangeetha-Bheeman: It is agreed that the suffix will be four letters(UPAB) plus one number representing the number of labels(1-9) - UPAB1

BenTopping (dismissed):
Looks good! I haven't tested the changes locally but I assume its working fine.

BenTopping (approved):

PR #945, opened 2022-11-01
Dpl 543 runs page bug

seenanair:
- Mapping value field in custom component
- Fall through attribute binding
-  Added unit test cases


stevieing (changes requested):
All seems good to me. Might be worth adding back in some comments about what kind of attributes can be passed along with a bit about how it works?

src/components/shared/TractionInput.vue
```diff
@@ -3,14 +3,9 @@
     <label v-if="title">{{ title }}</label>
     <input
       :id="id"
-      :value="content"
+      v-bind="$attrs"
```
BenTopping, line 6: Does this work by spreading/binding all passed in props to this component? If so does that mean we still need to explicitly declare :value ?

tests/unit/components/shared/TractionInput.spec.js
```diff
@@ -0,0 +1,73 @@
...
+        test: '',
+      },
+    })
+    const textInput = comp.find('input')
+    textInput.element.value = 'Test value'
+    textInput.trigger('input')
+    expect(comp.vm.test).toBe('Test value')
+  })
+
+  it('updates the input when changing the v-model value externally,', async () => {
```
BenTopping, line 60: Nice test!

BenTopping: Is the <div> surrounding <traction-input> needed?

seenanair:
> All seems good to me. Might be worth adding back in some comments about what kind of attributes can be passed along with a bit about how it works?

Sure, will do

src/components/shared/TractionInput.vue
seenanair: From the documentation I read, it says we need explicit binding for value field as this is a custom component wrapper.

tests/unit/components/shared/TractionInput.spec.js
seenanair: Removed 👍

seenanair:
> All seems good to me. Might be worth adding back in some comments about what kind of attributes can be passed along with a bit about how it works?

@stevieing  Comments added

src/components/shared/TractionInput.vue
BenTopping: 👍 Do you have a link to the doc?

seenanair: https://v2.vuejs.org/v2/guide/components.html#Using-v-model-on-Components
https://vuejs.org/guide/components/attrs.html
Also, if you search for Creating an Input component using Vue, you can find similar documentation - like https://dev.to/abhisheksarmah/creating-an-input-component-using-vue-and-tailwindcss-3a6b

seenanair:
@stevieing @BenTopping  I think I have made the changes you suggested
- Documentation added
- test case update
- removed id field prop as it will mapped as part of $attrs

stevieing (approved):

PR #947, opened 2022-11-02
DPL - 522 Remove references to PMB V1

Sangeetha-Bheeman:
Closes #

*
*
* ...


BenTopping (dismissed):
Not sure if its in scope but would be nice to change pmbV2 references to now just be pmb (remove v2 naming) e.g. in [here](https://github.com/sanger/traction-ui/blob/develop/src/store/printMyBarcode/actions.js) up to you though

tests/unit/store/index.spec.js
```diff
@@ -14,7 +14,6 @@ describe('index', () => {
     describe('api', () => {
       it('contains multiple resources', () => {
         expect(Store.state.api.traction).toBeDefined()
-        expect(Store.state.api.printMyBarcode).toBeDefined()
```
BenTopping, line 17: I think this should be tested still - if you remove the v2 from pmb api config name

src/api/Config.json
```diff
@@ -94,16 +94,6 @@
...
-    "apiNamespace": "v1",
-    "baseURL": "VITE_PRINTMYBARCODE_BASE_URL",
-    "resources": [
-      {
-        "name": "print_jobs"
-      }
-    ]
-  },
   {
     "name": "printMyBarcodeV2",
```
BenTopping, line 98: able to remove v2?

src/store/printMyBarcode/getters.js
```diff
@@ -1,6 +1,6 @@
 const getters = {
   sampleExtractionTubeRequest: (state, getters, rootState) => rootState.api.sampleExtraction.assets,
-  printJobV2Request: (state, getters, rootState) => rootState.api.printMyBarcodeV2.print_jobs,
+  printJobRequest: (state, getters, rootState) => rootState.api.printMyBarcodeV2.print_jobs,
```
BenTopping, line 3: if you change pmbv2 config name this will need updating to new name

BenTopping (changes requested):
Nice, I think that makes things a little cleaner in traction. Found one more v2 reference in api config if you could change that one over as well

BenTopping (approved):
Nice looks good!

PR #960, opened 2022-11-17
DPL-501 label printing - UAT comments done

Sangeetha-Bheeman:
Closes #

* Grouped suffix options under workflow in the suffix selection dropdown
* Made No. of labels optional
* Dropped the -1 from the suffix, if no of labels is 1
* Added - between the suffix and no of labels
* Removed the printers AA309bc2 and AA309bc3 from drop down as they are plate label printers


src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -153,16 +156,53 @@ export default {
     },
     suffixedBarcodes() {
```
stevieing, line 157: This method although working effectively could benefit from a refactor. I don't think that is necessary at this stage as it would be better to get it into UAT and we will make further changes when we add plate support.

I would however recommend adding some comments to explain what is going on as it is fairly complex.

src/config/SuffixList.json
```diff
@@ -1,86 +1,111 @@
 [
   {
```
stevieing, line 2: Great. This is much better as it will allow the flexibility to add more workflows.

stevieing (changes requested):

src/components/labelPrinting/LabelPrintingForm.vue
Sangeetha-Bheeman: Added some comments. Maybe will refactor later when adding plate support.

stevieing (approved):

PR #965, opened 2022-11-18
908 dpl 515 include labware source on labels

stevieing:
Closes #

- moved PacbioLibraryIndex to use createPrintJob and create labels within view
- moved PacbioPoolIndex to use createPrintJob and create labels within view
- moved PacbioSampleIndex to use createPrintJob and create labels within view (spurious as sample labels aren't printed and there is no barcode)
- moved SaphyrLibraryIndex to use createPrintJob and create labels within view 
- moved SaphySampleIndex to use createPrintJob and create labels within view (spurious as sample labels aren't printed and there is no barcode)
- added getCurrentDate helper
- added printLabelHelpers to create the data necessary for label printing. This should fix current bugs and allow for reuse
- moved labelPrintingForm to use printLabelHelpers and createPrintJob
- removed printHelper mixin
- removed printJob from actions and associated methods.
- removed consts lib and usage

Rationale:
- Everything was going through the printJob methods in printMyBarcode actions. There was some slightly strange behaviour and also this meant every label could only have set fields on it. This meant that modifying labels for any printing page was not possible. There were also some methods which were duplicated in the actions purely for testing. This may lead to false positives
- I have standardised the methods across the views which will allow for further refactoring if necessary.
- I have moved to a new method createPrintJob and label creation has moved to components which allows us to add different fields on each type of label
- The label printing page had some odd behaviour which was causing some bugs. These were difficult to detect as the method was not easily testable
- I moved this to the labelPrintingHelpers so I could work out how to construct the labels and test it. This will also promote reusability
- These changes will allow us to easily create different printing labels for ONT
- I removed the const library as it was adding a lot of overhead with little value. A lot of the constanst were not being used at all some files were using a mixture and some were only being used in a single file.

Still to do:
- I will go on site this week and deploy from branch to check that the printing works as requested
- I will look at renaming some of the methods and libraries to reflect the meaning better.
- Try and fix the ui issue where labels don't appear correctly on printing modal

@harrietc52 and @Sangeetha-Bheeman I have marked you both as code reviewers as you have both worked on this page. There is quite a lot to digest so let me know if you need any more info. Hopefully the info above should help.


src/views/saphyr/SaphyrLibraries.vue
```diff
@@ -156,16 +155,46 @@ export default {
...
+    */
+    createLabels() {
+      const date = getCurrentDate()
+      return this.selected.map(({ barcode }) => {
+        return {
+          barcode,
+          first_line: 'Saphyr - Library',
+          second_line: date,
+          third_line: barcode,
+          label_name: 'main_label',
```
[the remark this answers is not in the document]
stevieing: We would have to do some work in the service for Saphyr. It is not currently being used so is not a priority.

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -112,114 +95,74 @@
...
+      })
     },
-    onSubmit(event) {
+  },
+  methods: {
+    /*
+      Creates the print job and shows a success or failure alert
+      @param {event}
+    */
+    async printLabels(event) {
```
harrietc52, line 154: How come `event.preventDefault()` is needed?

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -112,114 +95,74 @@
...
-  barcode: null,
-  selectedSuffix: null,
-  selectedNumberOfLabels: null,
-  selectedPrinterName: null,
-  copies: null,
+  sourceBarcodeList: null,
+  suffix: null,
+  numberOfLabels: null,
+  printerName: null,
+  copies: '1',
```
harrietc52, line 111: Does this want to be an integer, rather than a string?

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -112,114 +95,74 @@
...
+
+      const { success, message = {} } = await this.createPrintJob({
+        printerName: this.form.printerName,
+        labels: this.labels,
+        copies: this.form.copies,
+      })
+
+      this.showAlert(message, success ? 'success' : 'danger')
+
+      return { success, message }
```
harrietc52, line 165: Does this need to be returned? If it's only added for the test maybe test the `showAlert` instead

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -142,7 +141,7 @@ export default {
         await this.setTags()
         this.tractionTags.forEach((tag) => this.tagOptions.push(tag.group_id))
       } catch (error) {
-        this.showAlert(consts.MESSAGE_ERROR_FIND_TAGS_FAILED + error.message, 'danger')
+        this.showAlert('Failed to find tags in Traction' + error.message, 'danger')
```
harrietc52, line 144: Maybe a space or a colon space after this message/could add a test for it too

harrietc52: How come removed the constants too?

src/store/printMyBarcode/actions.js
```diff
@@ -1,74 +1,56 @@
...
-  const request = getters.printJobRequest
-
-  const labelTemplateName = getters.tubeLabelTemplateName
-
-  const payload = createPrintJobJson(params, labelTemplateName)
+/*
+  TODO. The only action we have is createPrintJob
+  Question: Do we need this in the store at all?
+  The action does not relate to the store so is maybe better as a service.
+*/
```
harrietc52, line 7: Yep, would agree with this comment. Doesn't need to be in the store as it isn't handling common state which needs to be accessible throughout the application

src/store/printMyBarcode/actions.js
```diff
@@ -1,74 +1,56 @@
...
+    : // print my barcode does not return the errors in the correct format
+    // so we need to extract the errors and turn them into something meaningful
+    data?.errors?.length > 0
+    ? data.errors.map((e) => e.source.pointer + ' ' + e.detail).join(', ')
+    : // otherwise we can just return the errors
+      errors
 
-const trimBarcode = (suffix, barcode) => {
-  return suffix ? barcode.slice(0, barcode.lastIndexOf('-')) : barcode
+  return { success, message }
```
harrietc52, line 49: Nice, will reuse this 👍

src/views/pacbio/PacbioLibraryIndex.vue
```diff
@@ -171,9 +170,40 @@ export default {
...
+    },
+    /*
+      Creates the print job and shows a success or failure alert
+      @param {String} printerName The name of the printer to send the print job to
+    */
+    async printLabels(printerName) {
+      const { success, message = {} } = await this.createPrintJob({
+        printerName,
+        labels: this.createLabels(),
+        copies: '1',
```
harrietc52, line 202: Does this page support 1 copy only? And does it need to be a string?

tests/e2e/specs/visit_label_printing_page.cy.js
```diff
@@ -58,15 +56,4 @@ describe('Label Printing page', () => {
...
-    cy.intercept('/v2/print_jobs', {
-      statusCode: 422,
-      body: {
-        errors: [],
-      },
-    })
-    cy.get('#submit-button').click()
-    cy.contains('Unknown')
-  })
 })
```
harrietc52, line 59: Do you want to remove this test?

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -112,114 +95,74 @@
...
-    suffix() {
-      return this.form.selectedSuffix === 'No suffix' ? '' : this.form.selectedSuffix.slice(0, 4)
-    },
-    printerName() {
-      return this.form.selectedPrinterName
+    labels() {
+      const date = getCurrentDate()
+      const suffixItem = this.suffixItems[this.form.suffix]
+
+      // it is possible for there to be no barcodes so we need to add a guard
```
harrietc52, line 138: Maybe add a test for this guard

tests/unit/store/printMyBarcode/actions.spec.js
```diff
@@ -1,137 +1,81 @@
...
+        {
+          barcode: 'TRAC-1',
+          first_line: 'line 1',
+          second_line: 'line 2',
+          third_line: 'line 3',
+          fourth_line: 'line 4',
+          label_name: 'main_label',
+        },
+      ],
+      copies: 1,
```
harrietc52, line 20: `copies` here is an int, maybe change all to integers

tests/unit/store/printMyBarcode/actions.spec.js
```diff
@@ -1,137 +1,81 @@
...
-      }
+      const { success, message } = await createPrintJob(
+        { rootState, state: defaultState },
+        { ...rest },
+      )
 
-      const result = Actions.createPrintJobJson(params, tubeLabelTemplateName)
-      expect(result).toEqual(createPrintJobJson)
+      expect(success).toBeFalsy()
+      // looks like there is an issue with the errors again. How many times do we have to fix this.
```
harrietc52, line 77: What is wrong with it?

tests/unit/views/pacbio/PacbioSampleIndex.spec.js
```diff
@@ -1,6 +1,7 @@
 import Samples from '@/views/pacbio/PacbioSampleIndex'
 import { mount, localVue, store, Data, router } from '@support/testHelper'
 import Response from '@/api/Response'
+import { beforeEach, describe, expect, it } from 'vitest'
```
harrietc52, line 4: Maybe these could be pulled out to the testHelper so they don't have to be imported for every test

harrietc52 (commented):
Looks good, a few minor comments

src/config/SuffixList.json
```diff
@@ -3,108 +3,171 @@
...
-        "tubeStage": "Labelling",
-        "suffix": "BDLE"
+        "stage": "Labelling",
+        "suffix": "BDLE",
+        "text": "BDLE - Labelling",
+        "value": "BDLE",
+        "workflow": "BioNano"
       }
     ]
   }
```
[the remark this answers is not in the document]
stevieing: I did think about that but felt that the programmatic effort needed to then duplicate or create those fields added extra overhead. In this case you can just load stuff as is and there is also the flexibility to modify those fields.

src/consts/consts.js
```diff
@@ -1,37 +0,0 @@
...
-  MESSAGE_ERROR_GET_TRACTION_TUBES,
-  MESSAGE_ERROR_SINGLE_TYPE,
-  MESSAGE_ERROR_INVALID_BARCODES,
-  MESSAGE_ERROR_DELETION_FAILED,
-  MESSAGE_ERROR_FIND_TAGS_FAILED,
-  MESSAGE_ERROR_CREATE_LIBRARY_FAILED,
-  MESSAGE_ERROR_INTERNAL,
-  MESSAGE_WARNING_NO_BARCODES,
-  MESSAGE_ERROR_CREATE_RUN_FAILED,
-}
```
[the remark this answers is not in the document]
stevieing: I might have explained this in my rationale and again I was in 2 minds but my instinct told me it made sense to remove. It didnt seem to add much value and it felt like premature optimisation. Only 5 of the constants were being used and most of the rest were only being used in 1 file.

src/components/labelPrinting/LabelPrintingForm.vue
stevieing: Because it is a form so the submit button on a form the default would be to refresh the page or move to another one.

src/components/labelPrinting/LabelPrintingForm.vue
stevieing: Maybe. Not sure it makes a difference. I was thinking that we could remove it all together but wasn't sure if there were any side effects.

src/components/labelPrinting/LabelPrintingForm.vue
stevieing: Again I did think about that. My view was that it is a function so returning something should be standard and it felt more natural to test the return rather than something that was independent. Not sure either way.

tests/e2e/specs/visit_label_printing_page.cy.js
stevieing: Yes. Wasn't really clear on it's relevance. I have modified the code so that it does always return an error.

tests/unit/store/printMyBarcode/actions.spec.js
stevieing: I am just a bit unclear of why this is different to other API's. Print My Barcode is behaving differently I think?

tests/unit/views/pacbio/PacbioSampleIndex.spec.js
stevieing: These are not necessary as these are imported by default. VSCode adds them not me. I gave up removing them because it was happening every time I added any of those identifiers and I haven't tried to fix it.

src/components/labelPrinting/LabelPrintingForm.vue
stevieing: The current tests already cover it. Because it is reactive any test that does not have barcodes was failing which is why I added a guard. So the tests started failing and then I added the guard. Hope that makes sense?

BenTopping (approved):

PR #972, opened 2022-11-25
Custom pagination component with feature flag support

seenanair:
- Custom pagination component without any library using tailwind.
- Also supports feature flag 'enable_custom_pagination' 
     - If enabled, displays custom component, otherwise the bootstrap component


src/components/shared/TractionPagination.vue
```diff
@@ -0,0 +1,194 @@
...
+    //Get page numbers to display in page-buttons
+    pages() {
+      const range = []
+      for (let i = this.startPage; i <= this.endPage; i++) {
+        range.push({
+          name: i,
+        })
+      }
+      return range
+    },
```
stevieing, line 156: there is a nifty way where you can use Array.from. In this case it would be 
`Array.from({length: this.endPage - this.startPage}, (v, k) => k+this.startPage)`
Not necessary but cleaner maybe ...

seenanair: Sure, that's a good suggestion. I will do it

seenanair: @stevieing I have updated the review comments.

BenTopping:
Is there a branch/PR to add the feature flag in the service?

stevieing:
> Is there a branch/PR to add the feature flag in the service?

You can just go to flipper and add it. No need for a pr.

PR #979, opened 2022-12-06
Dpl 478

### LINKED ISSUE #885: DPL-479 - As a long read user I would like to be able to upload an extraction QC spreadsheet so that ToL can analyse qc results for reporting
**User story**
As a long read user I would like to be able to upload an extraction QC spreadsheet so that ToL can analyse qc results for reporting

Hermione / Esther

Hermione or other long read users

**Acceptance criteria**
- [x] A page where user can upload a csv file of an agreed format
- [x] A file upload input box
- [x] An upload button
- [x] When the button is pressed it will persist the records
- [x] Once the button is pressed it should be disabled
- [x] While the csv is being processed there should be visual feedback
- [x] Once processing is finished the button should be enable
- [x] If the upload is successful there should be some feedback indicating that it is a success along with an indication of how many qc results have been created
- [x] If the upload fails then there should be some feedback giving reasons as to why it has failed with an indication of where it has failed
- [x] There should be wiggle room if the use tries to upload previously recorded records. Rather than failing it will just create a new record. The qc results table is idempotent so should not overwrite amend or delete.

**Dependencies**
- sanger/unified_warehouse#397
- sanger/traction-service#815

**References**

- sanger/traction-ui#855

**Additional context**
Example spreadsheet to follow.

harrietc52:
Closes #885
Coupled with https://github.com/sanger/traction-service/pull/844


DPL-478-todo.md
```diff
@@ -0,0 +1,32 @@
+# Notes
```
BenTopping, line 1: Is this file still needed?

src/components/extractionQc/ExtractionQcForm.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <div>
+    <traction-form class="text-left" @submit="onSubmit">
+      <traction-sub-section title="Which QC Results would you like to upload?" class="py-6">
+        <traction-select
+          v-model="usedBySelected"
+          :options="usedByOptions"
+          :state="Boolean(usedBySelected) ? true : null"
```
BenTopping, line 8: state can be simplified to !!usedBySelected

src/components/extractionQc/ExtractionQcForm.vue
```diff
@@ -0,0 +1,74 @@
...
+          required
+        ></traction-select>
+      </traction-sub-section>
+      <traction-sub-section title="CSV File" class="py-6">
+        <traction-file
+          v-model="file"
+          :state="Boolean(file) ? true : null"
```
BenTopping, line 15: simplifies to !!file

src/components/extractionQc/ExtractionQcForm.vue
```diff
@@ -0,0 +1,74 @@
...
+const EXTRACTION_USED_BY = 'extraction'
+
+export default {
+  name: 'ExtractionQcForm',
+  mixins: [Api],
+  data() {
+    return {
+      usedByOptions: [
+        { value: null, text: 'Please select a option' },
+        { value: EXTRACTION_USED_BY, text: 'Extraction' },
```
BenTopping, line 44: not sure this needs to be abstracted to a const variable since the data is changed and its only used here

src/components/extractionQc/ExtractionQcForm.vue
```diff
@@ -0,0 +1,74 @@
...
+      await this.postCSV()
+    },
+    async postCSV() {
+      this.busy = true
+      try {
+        const csv = await this.file.text()
+        let data = { csv: csv, usedBySelected: this.usedBySelected }
+        await createQcResultsUploadResource(this.qcResultUploadsRequest, data)
+
+        // Todo: clear alerts
```
BenTopping, line 65: do you still plan on doing this?

src/components/extractionQc/ExtractionQcForm.vue
```diff
@@ -0,0 +1,74 @@
...
+        { value: null, text: 'Please select a option' },
+        { value: EXTRACTION_USED_BY, text: 'Extraction' },
+      ],
+      file: null,
+      usedBySelected: null,
+      busy: null,
+    }
+  },
+  computed: {
+    qcResultUploadsRequest: ({ api }) => api.traction.qc_results_uploads.create,
```
BenTopping, line 52: does this need to be done here and if so does it need to be computed? could it just be added directly into the QcResultsUpload file?

tests/e2e/specs/visit_home_page.cy.js
```diff
@@ -1,8 +1,20 @@
 // https://docs.cypress.io/api/introduction/api.html
 
 describe('My First Test', () => {
```
BenTopping, line 3: Probably change name of tests

tests/unit/services/traction/QcResultsUpload.spec.js
```diff
@@ -0,0 +1,108 @@
+import { createQcResultsUploadResource } from '@/services/traction/QcResultsUpload'
+
+describe('QcResultsUpload', () => {
+  describe('#createQcResultsUploadResource', () => {
+    // Setup some of the parameters we'll be testing with
+    let csv = `,,SAMPLE INFORMATION,,,,,,,,,,,,,VOUCHERING,,,,EXTRACTION/QC,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,COLUMN JUST FOR TOL,COLUMN JUST FOR TOL,SE LIMS,
```
BenTopping, line 6: maybe extract this and import an actual csv in tests/data/csv

tests/unit/services/traction/QcResultsUpload.spec.js
```diff
@@ -0,0 +1,108 @@
...
+            attributes: {
+              csv_data: csv,
+              used_by: usedBySelected,
+            },
+          },
+        },
+      })
+    })
+
+    // Error response currently thrown - to fix in Service
```
BenTopping, line 60: This still needed?

BenTopping (changes requested):
Few minor comments, i'd suggest adding a cypress test for the page as well

src/components/extractionQc/ExtractionQcForm.vue
```diff
@@ -0,0 +1,74 @@
+<template>
```
stevieing, line 1: Looking at this page in the ui and wondering if we need to change the language a bit? At the top it says extraction qc and then in the page it says Which QC Results would you like to upload? and then you have to select extraction qc.

src/components/extractionQc/ExtractionQcForm.vue
harrietc52: I can replace it with `!!(usedBySelected) ? true : null` if that's any nicer, but not `!!usedBySelected`, as we want to return null instead of false, otherwise the select appears red by default

src/components/extractionQc/ExtractionQcForm.vue
harrietc52: Same as https://github.com/sanger/traction-ui/pull/979#discussion_r1049757454

src/components/extractionQc/ExtractionQcForm.vue
harrietc52: Nope, will remove comment

tests/unit/services/traction/QcResultsUpload.spec.js
harrietc52: Nope, been fixed

src/components/extractionQc/ExtractionQcForm.vue
harrietc52: @BenTopping @stevieing I think it would defined in this file, unless I pass in the store object to the `createQcResultsUploadResource` function, so not sure that helps it much?
I could remove it from a computed, but the api function is defined as a computed prop in the Api module, so thought this was neater than `let qcResultUploadsRequest = this.$store.getters.api.traction.qc_results_uploads.create`. But don't mind

src/services/traction/QcResultsUpload.js
```diff
@@ -0,0 +1,30 @@
+import { handleResponse } from '@/api/ResponseHelper'
+/*
+  UPDATE
+*/
+
+const createQcResultsUploadResource = async (request, { csv, usedBySelected }) => {
+  const { success, data, errors } = await handleResponse(
+    // Why extra data {} needed ?
```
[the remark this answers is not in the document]
harrietc52: @stevieing I wasn't sure why the was needed
```
{
      data: {
        data: {
}}}
```

src/components/extractionQc/ExtractionQcForm.vue
harrietc52: Yes happy to, do you have any opinions on what would be the best wording for consistency?

DPL-478-todo.md
harrietc52: Removed

src/components/extractionQc/ExtractionQcForm.vue
harrietc52: Done

src/components/extractionQc/ExtractionQcForm.vue
harrietc52: I've left as is for now

tests/e2e/specs/visit_home_page.cy.js
harrietc52: Done

tests/unit/services/traction/QcResultsUpload.spec.js
harrietc52: Done

tests/unit/components/extractionQc/ExtractionQcForm.spec.js
```diff
@@ -0,0 +1,131 @@
+import ExtractionQcForm from '@/components/extractionQc/ExtractionQcForm'
+import { localVue, mount, store } from '@support/testHelper'
+import { describe } from 'vitest'
+import * as QcResultsUpload from '@/services/traction/QcResultsUpload'
+
```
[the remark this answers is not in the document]
harrietc52: Left for now, but if you're looking at the e2e test and think something is covered else where, feel free to remove it

src/components/extractionQc/ExtractionQcForm.vue
BenTopping: OK. Generally computed properties re-render quite a few times which would be unnecessary in this case but should be fine to leave in

stevieing (approved):

BenTopping (approved):

PR #992, opened 2022-12-08
DPL-581 bug pacbio pool warning icons

### LINKED ISSUE #975: DPL- 581 [BUG] Traction - Investigate error indication displayed on the 'Pools' edit screen
**Description**
An error indication (Exclamation with red highlighting) is displayed on the input field corresponding to the 'Tag' column on the 'Pools' edit screen.

image

**RT Ticket Number**
N/a

**To Reproduce**
1. Go to https://traction.psd.sanger.ac.uk/#/dashboard > Pools > Pools Tab > Edit
2. Navigate to the Pooled Samples section > 'Tag' value input field

**Actual behaviour**

1. The error indication is displayed at all times.

**Expected behaviour**

1. The error indication should only be displayed if applicable else, it should not be displayed.

**Additional context**
N/a

andrewsparkes:
Closes #975


stevieing (dismissed):
All seems fine to me. Definitely think Seena should check as well.

src/components/pacbio/PacbioPoolLibraryEdit.vue
```diff
@@ -110,9 +116,20 @@ export default {
...
       default: false,
     },
+    validated: {
+      type: Boolean,
+      default: false,
+    },
+    notify: {
+      type: Function,
+      required: true,
+    },
```
seenanair, line 126: Probably better to add comments on 'validated' field where it is set, how it is used and also on what notify function does

src/components/pacbio/PacbioPoolLibraryEdit.vue
```diff
@@ -146,18 +163,35 @@ export default {
...
+        return true
+      } else {
+        return !this.validationRequiredFields[attribute]
+      }
+    },
+    setValidationRequired() {
+      if (this.validated) {
+        this.validationRequiredFields = []
+      }
+    },
```
seenanair, line 184: Do you think the method name is bit confusing? It is actually resetting all flags related to fields , probably something like resetAllValidationFields ?

src/components/pacbio/PacbioPoolEdit.vue
```diff
@@ -163,13 +161,14 @@ export default {
         this.parsedFile = false
       }
     },
+    onFieldUpdate() {
+      this.validated = false
+    },
```
seenanair, line 166: Maybe, comment here as well  on what it does would be helpful  for future support

seenanair:
e2e test cases are missing these warning checks, I think. So, probably it might be useful to include those in the test cases so it won't slip any more issues related to this into production in future

src/components/pacbio/PacbioPoolLibraryEdit.vue
andrewsparkes: Yes agreed, done

src/components/pacbio/PacbioPoolEdit.vue
andrewsparkes: Yes, done

seenanair (approved):
Great, all looking good 👍

stevieing (approved):

PR #1025, opened 2023-01-12
General reception bug fix

BenTopping:
Prevented alerts from growing over the page, fixed TractionFieldGroup attribute bug.

Alert contains now has a max height
Alerts auto disappear after 15 seconds
General reception input fields now accept data properly

### DIFF
--- a/src/components/shared/TractionFieldGroup.vue
+++ b/src/components/shared/TractionFieldGroup.vue
@@ -37,7 +37,6 @@
       :is="component"
       v-if="component"
       :id="fieldId"
-      v-model="getAttribute"
       :value="value"
       :data-attribute="attribute"
       v-bind="componentProps"
@@ -69,12 +68,6 @@ export default {
     value: { type: [String, Number, Object, Array], required: false, default: null },
   },
   data: (component) => ({ fieldId: component.for || uniqueId() }),
-
-  computed: {
-    getAttribute() {
-      return this.attribute
-    },
-  },
   methods: {
     input(value) {
       // Support either native components emiting events, or components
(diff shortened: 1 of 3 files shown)


stevieing (approved):

src/components/shared/TractionFieldGroup.vue
```diff
@@ -37,7 +37,6 @@
       :is="component"
       v-if="component"
       :id="fieldId"
-      v-model="getAttribute"
```
seenanair, line 40: v-bind="$attrs" Is this another way of doing it , where we can avoid getAttribute computed method. But  please check which is the better way as I am not sure which is more appropriate.

BenTopping: I think you know more than me in this area so happy to go with whatever you think. The issue with this v model method was that it couldn't 'set' values because it was a computed method. But not sure if its needed at all. If we bind attrs can we remove the value prop as well or does it just replace value?

seenanair: As my understanding v-bind="$attrs" is like fall through attribute so we don't need to  explicitly declared in the receiving component's props especially in cases where we need all props to bind directly to the wrapped component. For e.g if a button is wrapped in div and we want all fallthrough attributes like class and v-on listeners to be applied to the inner <button>, not the outer <div>. We can achieve this with inheritAttrs: false and v-bind="$attrs"

seenanair (approved):
@BenTopping I tried it , but it is not getting initialized when default values are given in props using v:bind:$attrs. So, the way you have implemented is good. 👍

PR #1033, opened 2023-01-19
DPL-558 Classify the barcodes based on their relevant tag sets when creating a new sample library

yoldas:
Closes #

- [x] Add tag set selection input (first dropdown)
- [x] Add tag select input (second dropdown)
- [x] Filter tags based on selected tag-set 
- [x] tag-set change should reset selected library tag
- [x] reuse fetchPacbioTagSets and createLibraryInTraction [Not reusing fetchPacbioTagSets anymore]
- [x] Use tag id instead of searching it by group_id  
- [x] Fix e2e test create_pacbio_library_from_samples_page
- [x] Fix createLibraryInTraction unit test
- [x] Include pacbio tagSets only 
- [x] Disable tag selection if no tag-set is selected 
- [x] Create a new fetchPacbioTagSets in libraries instead of using from poolCreate
- [x] Use getters instead of state mapping 
- [x] Store computed tagSet and tag options in state
- [x] selected tag id is mapped to library.tag.id


yoldas:
- [] Add tag-set-select input
- [] Add tag select input

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -22,10 +22,20 @@
...
+          <traction-select
+            id="tag-set-input"
+            v-model="selectedTagSetId"
+            data-type="tag-set-list"
+            :options="tagSetOptions"
+            @input="resetSelectedTagId"
+          ></traction-select>
+        </traction-form-group>
+
+        <traction-form-group id="tag-select-input" label-for="tag-input">
```
BenTopping, line 35: Would be nice to have this conditionally disabled based on if a tag set has been selected

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -125,26 +135,57 @@ export default {
...
-      tagOptions: [{ value: '', text: 'No tag' }],
+      library: { tag: { id: '' }, sample: {} },
+      selectedTagSetId: '',
+      selectedTagId: '',
     }
   },
   computed: {
-    ...mapGetters('traction', ['tractionTags']),
+    // Map fetched Pacbio tag sets and tags
+    ...mapState('traction/pacbio/poolCreate', {
```
BenTopping, line 145: I see the ability to reuse VueX methods here and that is something we can look at in the future (implementing something similar for ONT at the moment). But for the pacbio VueX stores I would avoid taking methods/state across modules because that makes a dependency between them. In this instance it may cause side effects to the tag data used in the poolCreate page. I would just copy the vueXmethods from pacbio/poolCreate into pacbio/libraries.

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -125,26 +135,57 @@ export default {
...
+      tagSets: (state) => state.resources.tagSets,
+      tags: (state) => state.resources.tags,
+    }),
+    // Return options for the first dropdown
+    tagSetOptions() {
+      const tagSets = Object.values(this.tagSets)
```
BenTopping, line 151: I would use a getter to get TagSets/tags instead of map state in the component.

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -125,26 +135,57 @@ export default {
...
+      const options = tagSets.map(({ id: value, name: text }) => ({ value, text }))
+      const placeholder = { value: '', text: 'Please select a tag set' }
+      return [placeholder, ...options]
+    },
+    // Return options for the second dropdown
+    tagOptions() {
+      const placeholder = { value: '', text: 'Please select a tag' }
```
BenTopping, line 158: You could do this formatting in the tag fetch when the tag data is returned form the service. E.g tagset state in would look like tagSets: { 'tag_set_example': {  tags: [ {'id': 1, 'group_id' : '123'} ]}}.

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -22,10 +22,20 @@
...
+            :options="tagSetOptions"
+            @input="resetSelectedTagId"
+          ></traction-select>
+        </traction-form-group>
+
+        <traction-form-group id="tag-select-input" label-for="tag-input">
           <traction-select
             id="tag-input"
-            v-model="library.tag.group_id"
+            v-model="selectedTagId"
```
BenTopping, line 38: why did this need changing?

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: If we disable it, they can't change it again before submitting.

BenTopping (changes requested):
Think its nearly there. Couple comments about VueX usage

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: I will create a separate vuex module libraryCreate and do them separately.

src/components/pacbio/PacbioLibraryCreate.vue
BenTopping: Im not sure what you mean. Conditionally disabling the tag select based on whether a tagSet has been selected should be fine?

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: I explicitly separated the placeholder to make it visible. tagOptions depends on the selectedTagSet in the first dropdown. It has to be computed again.

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: It is for changing the logic in createLibraryInTraction. Instead of querying again in there to find the tag.id from tag.group_id, I directly used the Id of the tag.

src/components/pacbio/PacbioLibraryCreate.vue
sdjmchattie: Just to clarify what I think @BenTopping is saying.  If they haven't selected a tag set in the drop down above, then the specific tag drop down this comment is against should be disabled as it makes no sense to select a tag without selecting a tag set.  Then when a tag set has been selected, the specific tag drop down should become enabled.

yoldas: Now the second dropdown (tag) is enabled if there is a selection in the first (tagset). It will switch between enabled/disabled accordingly.

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: Now LibraryCreate has its own fetch to retrieve tagsets and tags and it writes computed options to its own state. I have replaced mapState with mapGetters as well.

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: I have replaced mapState with mapGetters to get options from the state.

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: Options are formatted and written to state. The component retrieves the options and adds the placeholders only, which can be customised easily there.

src/components/pacbio/PacbioLibraryCreate.vue
yoldas: I have changed this line back to library.tag.id . Note that we are not using group_id as it is not necessary. The previous code had another step to find id from group_id in createLibraryInTraction. We don't need that anymore as the id is available.

src/components/pacbio/PacbioLibraryCreate.vue
BenTopping: I dont think it particularly matters where the options are computed, i was more talking about how the tag set and tags data looks. I think the changes you made are good though and easy enough to change in the future if we need to

tests/e2e/specs/pacbio/create_pacbio_library_from_samples_page.cy.js
```diff
@@ -19,6 +19,7 @@ describe('Pacbio library creation from sample', () => {
     cy.get('#samples-table').contains('td', '5')
     cy.get('#samples-table').first().click()
     cy.get('#pacbioLibraryCreate').click()
+    cy.get('#tag-set-input').select('Sequel_16_barcodes_v3')
```
BenTopping, line 22: a check to see if tag-input is disabled before selecting tag set may be a good thing to test here.

BenTopping (approved):
Works well locally, looks good 👍

PR #1038, opened 2023-01-24
Disable QC Upload button after upload

harrietc52:

After uploading Qc Result CSV..
* The successful message or the error message is shown
* The message will still automatically disappear after 15 seconds
* The Upload File  button will be disabled after uploading (failed or successful)
* You can click Re-enable  to make the Upload File button enabled again, or you can refresh the page.

Non-blocking relationship with: https://github.com/sanger/traction-service/pull/900

### DIFF
--- a/src/components/qcResultsUpload/QcResultsUploadForm.vue
+++ b/src/components/qcResultsUpload/QcResultsUploadForm.vue
@@ -37,7 +37,8 @@
         size="lg"
         :disabled="!disableUpload"
         @click="disableUpload = !disableUpload"
-      >Re-enable</traction-button>
+        >Re-enable</traction-button
+      >
     </traction-form>
   </div>
 </template>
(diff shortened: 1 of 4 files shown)


stevieing (approved):

BenTopping (changes requested):
I made the alert auto disappear recently because some pages chat a lot of alerts at once and having to manually remove them all is a pain. I imagine QC needs them to stay for a while so that the users can read all the individual errors? Maybe keep alerts as auto disappearing but make a way to override the auto disappearing for use cases like this

src/components/qcResultsUpload/QcResultsUploadForm.vue
```diff
@@ -21,11 +21,24 @@
           required
         ></traction-file>
       </traction-sub-section>
-      <traction-button id="upload-button" type="submit" theme="create" size="lg" :disabled="busy">
+      <traction-button
+        id="upload-button"
+        type="submit"
+        theme="create"
+        size="lg"
+        :disabled="disableUpload"
```
BenTopping, line 29: I dont know the background of the issue but can you use the busy variable instead of disableUpload? i.e disabled (busy) whilst the file is processing and enabled (not busy) after processing.

BenTopping (approved):

Issue #941, opened 2022-11-01
DPL-433-3 - As a developer I need to reduce the technical debt in pacbio runs as it is not setup correctly which is causing hidden bugs and issues to ensure we are Revio ready


stevieing:
**User story**
As a developer I need to reduce the technical debt for pacbio runs as it is not setup correctly which is causing hidden bugs and issues. It is also not tested thoroughly or correctly in places which is making it difficult to debug. This is reducing confidence. There is also intermittent failures in the ci Cypress test. This is the only test that fails regularly.

Note: This work will also better prepare us for Revio. 

**Who are the primary contacts for this story**
Steve

**Who is the nominated tester for UAT**
PSD / Long Read

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Reorganise the vuex store so it is structured correctly
- [ ] Ensure that any asynchronous methods that are using promises handle errors correctly to ensure better user feedback
- [ ] Remove any false positive tests and replace them with working ones
- [ ] Reorganise tests to ensure that they are testing the correct things including removing any hidden asynchronous timeouts
- [ ] Move PacbioRun.js into store and reorganise it.
- [ ] Add appropriate end to end tests and modify existing ones
- [ ] Ensure that communication between store and components are correct
- [ ] Remove any object reassignment or passing objects between e.g. functions should have no side effects
- [ ] Remove any code that could lead to memory leaks e.g. Run.build receives an object and then returns that object. It should make a copy of that object before returning.
- [ ] Ensure that any promises are handled correctly
- [ ] Ensure that any functions that involve too much recursion are modified e.g. deserialize
- [ ] Add an appropriate level of documentation.
- [ ] Ensure that run creation and editing performs correctly
- [ ] Add a wrapper function/component to setup all data in store i.e. replace provide with data fetchers
- [ ] Create example payload so that front end and back end stories can be worked on independently

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
It will also be appropriate to change the service so that run creation and editing is done in a single request.

stevieing:
Tasks to complete. Add as necessary:

- [ ] modify editRun action to use handleResponse and provide better feedback.

SujitDey2022:
@stevieing copying in the enhancements discussed with James Watt

[...]

PR #1061, opened 2023-02-08
990 dpl 592 traction as a user i want to be able to search the sample libraries so that i can add the relevant samples to the wells on the run screen

enegoktan:
Closes #

* Users are able to search for the pools they want while creating and editing runs.
*
* ...


dpl-592-todo.md
```diff
@@ -0,0 +1,87 @@
+DPL-592
```
stevieing, line 1: Guess you can remove this file now?

src/store/traction/pacbio/runCreate/getters.js
```diff
@@ -9,4 +17,34 @@ export default {
   defaultSmrtLinkVersion: (state, getters) => {
     return Object.values(getters.smrtLinkVersionList).find((version) => version.default)
   },
+
+  // copied over from pools/getters.js
```
stevieing, line 21: If we copied this over verbatim then it sounds like as as part of the refactor we should reuse this function.

src/store/traction/pacbio/runCreate/mutations.js
```diff
@@ -1,11 +1,61 @@
...
+    const poolsObject = pools.reduce(function (acc, cur) {
+      acc[cur.id] = cur
+      return acc
+    }, {})
+    Vue.set(state, 'pools', {
+      ...state.pools,
+      ...poolsObject,
+    })
+  },
+  setTubes(state, tubes) {
```
stevieing, line 24: There are four functions all using the same structure. You could create a helper function which each one could then call.

tests/unit/components/pacbio/PacbioRunWellItem.spec.js
```diff
@@ -177,7 +179,7 @@ describe('Well.vue', () => {
 
     beforeEach(() => {
       newBarcode = 'TRAC-2-1'
-      store.state.traction.pacbio.pools = storePools
+      // store.state.traction.pacbio.runCreate = storePools
```
stevieing, line 182: can this be removed?

tests/unit/views/pacbio/PacbioRunShow.spec.js
```diff
@@ -191,6 +192,27 @@ describe('Run.vue', () => {
     })
   })
 
+  describe('#created', () => {
+    it('"findPools" gets called with a list of barcodes', () => {
```
stevieing, line 196: It might make sense to skip this with a reason so we know it is there. As it stands the test will pass.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -355,6 +355,7 @@ export default {
     },
     async updatePoolBarcode(row, barcode) {
       const index = row.index
+      await this.$store.dispatch('traction/pacbio/runCreate/findPools', { barcode: barcode })
```
BenTopping, line 358: If they manually type in a pool barcode in a well this method will be called every time they type a character. I do not think this will be noticeable from a user point of view but something to bear in mind.

BenTopping: Following on from that I actually changed this behaviour [here](https://github.com/sanger/traction-ui/pull/1056/files#diff-f4defb31124e9cc0b7c5d31a5d29574579c504f6c046647a068e411a12bd8b11R190). Worth pulling develop back into branch and checking it still behaves ok.

src/components/pacbio/PacbioPoolTubeItem.vue
```diff
@@ -6,6 +6,9 @@
     @dragstart="drag(barcode, $event)"
     @click="expanded = !expanded"
   >
+    <div class="flex justify-end content-end items-end">
```
BenTopping, line 9: I dont think `content-end` or `items-end` do anything here positionally so id remove them unless there is a specific reason.

src/components/pacbio/PacbioPoolSelectedList.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <div data-type="selected-pool-list">
+    <div class="flex flex-col py-3">
+      <div class="pools">
+        <traction-list-group class="list-group">
```
BenTopping, line 5: I would move the css classes at the bottom of the file into tailwind classes here. Then you can remove the list-group name.

src/store/traction/pacbio/runCreate/actions.js
```diff
@@ -1,4 +1,47 @@
...
+    commit('setRequests', requests)
+    return { success, errors }
+  } else {
+    return { success: false, errors: [`Unable to find pool with barcode: ${filter['barcode']}`] }
+  }
+}
+
+const actions = { findPools }
+
+export { findPools }
```
BenTopping, line 44: Exporting is a bit weird in this file. I think you can remove like 42 & 44 and move the find pools function below into `export default { ... } `

src/views/pacbio/PacbioRunShow.vue
```diff
@@ -133,6 +134,8 @@ export default {
         this.newRun()
       } else if (!this.newRecord) {
         await this.editRun(parseInt(this.$route.params.id))
+        const barcodes = this.poolBarcodes
+        await this.$store.dispatch('traction/pacbio/runCreate/findPools', { barcode: barcodes })
```
BenTopping, line 138: Nice way of loading the pools for existing runs 👍

BenTopping: If the pools IDs are available it may be faster to grab the pools based on their Ids but I don't think it will make much of a difference.

tests/unit/components/pacbio/PacbioPoolList.spec.js
```diff
@@ -20,10 +20,6 @@ describe('pacbioPoolList', () => {
     expect(wrapper.find('.pools')).toBeDefined()
   })
 
-  it('contains the correct data', () => {
```
BenTopping, line 23: Is there a reason this test was removed?

BenTopping (changes requested):
Looks really good, should make the user's life a lot easier!
Couple minor things to check but overall looks good 👍

tests/unit/components/pacbio/PacbioPoolList.spec.js
enegoktan: Yes I've actually moved this functionality to a new component PacbioPoolSelectedList but I'm realising I haven't tested this separately in there. I'll add a new spec test for that component to add this in instead?

BenTopping: OK yeah that sounds sensible

src/store/traction/pacbio/runCreate/actions.js
```diff
@@ -1,4 +1,47 @@
...
+      tags: 'group_id',
+      libraries: 'request,tag,run_suitability',
+    },
+    filter,
+  })
+  const response = await handleResponse(promise)
+
+  const { success, data: { data, included = [] } = {}, errors = [] } = response
+  // success is true with an empty list when no pools match the filter
+  if (success && data.length > 0) {
```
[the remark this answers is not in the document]
enegoktan: This will (hopefully) be resolved during the refactoring

tests/unit/components/pacbio/PacbioPoolList.spec.js
enegoktan: I'm now testing this functionality in the e2e test so will be removing the whole test file

src/components/pacbio/PacbioPoolSelectedList.vue
enegoktan: I've left this in for now as it seems to be connected to some other styles at the moment

src/store/traction/pacbio/runCreate/getters.js
enegoktan: This will also hopefully be resolved during the refactoring https://github.com/sanger/traction-ui/issues/941

stevieing (approved):

PR #1088, opened 2023-03-06
clearPoolData in pacbio runs

enegoktan:
Closes #

* clears the runCreate store from previous data when creating a new run
*
* ...


stevieing (approved):
Looks quite clean and you have also removed a method!

PR #1091, opened 2023-03-08
990-dpl-592-traction UAT fixes

enegoktan:

* Added search on enter
* Remain the cursor in the input after a search

### DIFF
--- a/src/components/LabwareFinder.vue
+++ b/src/components/LabwareFinder.vue
@@ -2,13 +2,19 @@
   <div class="flex flex-row">
     <traction-input
       id="labware-finder-input"
+      ref="search"
       v-model="searchValue"
       type="search"
       placeholder="Type to Search"
       label="Search value"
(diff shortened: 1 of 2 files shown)


stevieing (approved):
Nice.

src/components/LabwareFinder.vue
```diff
@@ -58,6 +60,8 @@ export default {
           this.showAlert(res.errors, 'danger')
         }
       })
+      // redirecting the focus to the input after searching
+      this.$refs.search.$refs.inputRef.focus()
```
yoldas, line 64: Shouldn't be this.$refs.inputRef.focus() ?

enegoktan: I think I need to reference _search_ first (finds the traction-input component that LabwareFinder uses) to be able to reference _inputRef_ which finds the LabwareFinder

PR #1092, opened 2023-03-09
Add default cost code to general reception

stevieing:
Closes #

*
*
* ...


stevieing:
> Looks fine. I know we have this value set in config in the service but I guess it doesnt make sense to make a call just to pick up the default cost code. Only suggestion is maybe make it a const instead of hard coding.

It gets set on create. We don't need to pull it back.

PR #1094, opened 2023-03-16
Rebased cleanup

BenTopping:
James G:
"We have a large amount of code which is unused once
flags have flipped. This change removes it all."


__mocks__/swrv.js
```diff
@@ -3,12 +3,8 @@ const baseURL = import.meta.env.VITE_TRACTION_BASE_URL
 const flags = {
   flipper_id: 'User',
   features: {
-    dpl_277_enable_general_reception: { enabled: true },
-    dpl_478_enable_qc_results_upload: { enabled: true },
     enable_custom_select: { enabled: true },
```
BenTopping, line 6: I am not sure whether we want to keep these flags or not. @seenanair any thoughts about when we should remove the custom component flags?

src/router.js
```diff
@@ -64,13 +61,7 @@ export default new Router({
       component: SaphyrView,
       meta: { pipeline: 'Saphyr' },
       children: [
-        { path: '', redirect: 'reception', meta: { page: 'Reception' } },
-        {
-          path: 'reception',
```
BenTopping, line 69: We are removing saphyr reception and currently there isnt another way to import into saphyr. Looking at prod it appears saphyr hasnt been used since 2019. I think if/when they want to start using it again we should just implement it as an option in the general reception?

src/views/GeneralReception.vue
```diff
@@ -1,159 +1,157 @@
...
-          </traction-muted-text>
-          <traction-select
-            id="sourceSelect"
-            v-model="source"
-            class="inline-block w-full"
-            :options="receptions"
-            data-type="source-list"
-          />
-        </div>
+  <div class="flex flex-row space-x-12 px-12">
```
BenTopping, line 2: The file changes look ugly but I the only difference is the removal of the flagged-feature wrapper and then subsequent indentation changes.

__mocks__/swrv.js
stevieing: I would say remove them. They are all in production.

src/router.js
stevieing: Agreed. I will check with SciOps to see what they want to do about it. I also suspect because of code rot there will be things that aren't working.

stevieing (dismissed):
Awesome. Nearly 2500 lines of code gone.

stevieing (approved):
All good. We will need to merge into DPL-433-3. I have checked with Emma about Saphyr but don't want to wait to merge this in.

PR #1110, opened 2023-04-04
DPL-705 Prevent pacbio pulling back all records

### LINKED ISSUE #1021: DPL-705 - As a developer I would like to change pacbio from pulling all records back as it is causing increasingly slow page times
**User story**
For performance reasons we should change all areas in pacbio that pull all records back to a solution where we apply a filter and search using the backend.

This will require pacbio pool/new to implement a service search for finding plates and tubes instead of pulling all records back and filtering in the front end.

This has already been done for ONT so we can use that logic. We also need to do this quickly as we have had to remove the pagination from the samples page which is causing serious performance degredation for one of the most used pages.

Ben T

N/A

**Acceptance criteria**
- [ ] Pacbio pools/new - Plates and tubes should be changed to only pull back records searched by the user  
- [x] Pacbio run edit/create - Pools list should be changed to only pull back records searched by the user (fix in #990)
N.B This work involves paginated the json api resources for pacbio plates, tubes and pools.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1021 

* Added findPacbioPlate and findPacbioTube to pacbio poolCreate actions
* Changed the way plates and tubes are searched for on the pacbio pool/new page
* Updated unit tests

This PR is coupled with https://github.com/sanger/traction-service/pull/974


seenanair:
@BenTopping  Just to let you know that the problem I told yesterday about rows not appearing was due to one of the changes we made. All sorted - e2e and unit test cases added

BenTopping:
@seenanair Thanks for you help on this, looks good!

seenanair:
> @seenanair Thanks for your help on this, looks good!

It is mutual, great help from you too. Much appreciated 👍

sdjmchattie (approved):

PR #1123, opened 2023-04-14
Dpl 701 traction form

### LINKED ISSUE #1098: DPL-701 Custom form component to replace b-form
**User story**
As part of clean migration from bootstrap (DPL-483) , create a custom component  to replace the bootstrap b-form 

Steve/Seena

**Acceptance criteria**
- [ ] The new component should contain all existing functionalities of b-form
- [ ] Support all existing features of b-form
- [ ] Compatibility with bootstrap props
- [ ] Feature flag support
- [ ] Replace all existing b-form references with new component
- [ ] Remove b-form from globally imported list

References
- #827

seenanair:
Closes #1098 

New custom component traction-form to replace b-form
Support for b-form using 'enable_custom_form' feature flag


src/components/shared/TractionForm.vue
```diff
@@ -0,0 +1,49 @@
...
+  <flagged-feature name="enable_custom_form">
+    <template #disabled
+      ><b-form :class="classes" @submit.prevent="submitAction" @reset.prevent="resetAction">
+        <slot />
+      </b-form>
+    </template>
+    <form
+      v-bind="$attrs"
+      :class="classes"
+      action=""
```
BenTopping, line 11: Can action/reset be removed since they are empty props or are they needed to override default behaviour

BenTopping (approved):
Looking good. Just to let you know I committed a change to deal with the one use case of [bootstrap form inlines](https://bootstrap-vue.org/docs/components/form#inline-form)

src/components/shared/TractionForm.vue
seenanair: Yes, that was my understanding that it is is needed to override the default behaviour.

seenanair:
> Looking good. Just to let you know I committed a change to deal with the one use case of [bootstrap form inlines](https://bootstrap-vue.org/docs/components/form#inline-form)

Great, thank you 👍

stevieing (approved):

PR #1124, opened 2023-04-18
Initial changes for replacing b-card

Sangeetha-Bheeman:
Closes #

*
*
* ...

### DIFF
--- a/src/components/shared/TractionCard.vue
+++ b/src/components/shared/TractionCard.vue
@@ -1,0 +1,12 @@
+<template>
+<div class="border-2 rounded border-white bg-white">
+  <slot />
+</div>
+</template>
+
+<script>
+
+export default {
+  name: 'TractionCard',
+}
+</script>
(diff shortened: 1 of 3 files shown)


seenanair (commented):
Looks good 👍 
Few things might be good to add 
1) a prop to accept tailwind styling classes as there are a couple of places it is called with extra styling
2) May be an existing bootstrap card support using a feature-flag? I believe bootstrap card component  is basically a wrapper with certain styling , So @stevieing do you think it is required in this case? Also  visual inspection in each page where it is called and compare it with it's equivalent (using bootstrap)  would be good to ensure the styling is intact with how it is now with bootstrap

stevieing (changes requested):
I don't think you need to create a new component for this. I have just changed the components to a standard div and there didn't seem to be any difference. Creating specialist components is only necessary when there are functionality changes. In cases like this it is a good idea to check first by regressing to a standard tag to see the differences.

stevieing (approved):

PR #1125, opened 2023-04-18
Removed Smrt Link v10 options.

stevieing:
Closes #

*
*
* ...


src/components/pacbio/PacbioRunWellDefaultEdit.vue
```diff
@@ -205,12 +167,12 @@ export default {
         '24.0',
         '30.0',
       ],
-      generateHifiOptions: {
```
BenTopping, line 208: If we are only supporting Sequel IIe now we also need to remove the ability to select it [from here](https://github.com/sanger/traction-ui/blob/develop/src/components/pacbio/PacbioRunInfoEdit.vue#L112)

BenTopping (dismissed):
Looks good. Searching v10 across the code base comes up with a few test files that mock the v10 smrtlink version so may be worth also removing those.

BenTopping (approved):

Issue #1105, opened 2023-03-31
DPL-719 - As a user I need to be able to create the default well options for SMRT Link 12 for Revio so that we can have the options prepopulated for each well


stevieing:
**User story**
As a user I need to be able to create the default well options for SMRT Link 12 for Revio so that we can have the options prepopulated for each well. This means that the users do not have to duplicate the well options each time so creates a considerable time saving.

**Who are the primary contacts for this story**
Steve / Craig / James

**Who is the nominated tester for UAT**
James / Craig

**Acceptance criteria**
To be considered successful the solution must allow for the ability to set a default option for:
- [ ] movie acquisition time (hours) - options same as Movie Time
- [ ] Pre-extension time (same values as per version 11)
- [ ] include Base Kinetics -yes or no
- [ ] Polmerase Kit - valid barcode

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

PR #1129, opened 2023-04-19
DPL - 716,719 - Ability to select v12 revio and v12 well defaults

### LINKED ISSUE #1104: DPL-716 - As a lab user when I create a run I should be able to select SMRT LInk Version 12 for Revio
**User story**
As a lab user when I create a run I should be able to select SMRT LInk Version 12 for Revio

Steve / Craig / James

Craig / James

**Acceptance criteria**
- [ ] When I visit the new run page I should see an option for SMRT Link 12 for Revio

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1104 #1105

* Reworked pacbio run well default edit to load fields dynamically


src/store/traction/pacbio/runCreate/run.js
```diff
@@ -44,9 +44,166 @@ const defaultWellAttributes = () => {
...
     demultiplex_barcodes: onInstrument,
     include_fivemc_calls_in_cpg_motifs: 'Yes',
+    movie_acquisition_time: null,
+    include_base_kinetics: null,
+    library_concentration: null,
+    polymerase_kit: null,
   }
 }
 
+const smrtLinkVersionDefaultComponents = {
```
BenTopping, line 54: Feel like it might be more sensible to have this data in the component its used it instead of here.

BenTopping:
> Code looks file but I have tried running it locally. Options change as expected but movie time options don't appear. Also there is no option to select Revio as a Systyem name which might be an idea to add for this story.

Movie times should be there. I had a typo which meant it showed up as Move time which may be where the confusion is. Also system name is pulled from the service so it may be you dont have 12_revio locally or that you dont have it set as active?

stevieing:
> > Code looks file but I have tried running it locally. Options change as expected but movie time options don't appear. Also there is no option to select Revio as a Systyem name which might be an idea to add for this story.
> 
> @stevieing Movie times should be there. I had a typo which meant it showed up as Move time which may be where the confusion is. Also system name is pulled from the service so it may be you dont have 12_revio locally or that you dont have it set as active?

I was thinking about system name not SMRT Link Version. We have added it to the back end but not the ui.

BenTopping:
Ah my mistake. I have added it.

stevieing (changes requested):
For me movie acquisition time still does not have the relevant options? Maybe I am missing something. Also wondering if Include Base Kinetics needs a default but that could wait until UAT.

stevieing (approved):

PR #1135, opened 2023-04-21
Fixed table data error not showing up latest records on multiple pages

seenanair:
- Default sort options in tableHelper mixin
- Initial data will be sorted by mixin if sort options given

Closes #

*
*
* ...

### DIFF
--- a/src/mixins/TableHelper.js
+++ b/src/mixins/TableHelper.js
@@ -2,6 +2,7 @@
  * A mixin to store commonly used functionality for the bootstrap table component
  */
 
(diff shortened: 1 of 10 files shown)


seenanair:
@stevieing @BenTopping This is the fix for table data not showing the latest when multiple pages given. I believe the only option left for us in this case is to manage the initial data in mixin itself , as the pagination and table are separate components and it is better to avoid any communication between them. So, I added sort option in mixin.
- Default sort options in tableHelper mixin
- Initial data will be sorted by mixin if sort options given

PR #1142, opened 2023-04-26
Consolidated pacbio run well components config

BenTopping:


src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -391,6 +206,25 @@ export default {
         this.localPools.push({ id, barcode: pool.barcode })
       })
     },
+    handleCustomProps(component) {
+      if (component.name == 'loading_target_p1_plus_p2') {
+        return {
+          ...component.props,
+          // This doesn't work at the moment as traction-input doesnt have a formatter prop
+          formatter: this.formatLoadingTargetValue,
```
BenTopping, line 214: @seenanair Maybe one to look at on a Friday. The idea here is that you can pass a formatter function into an input field and it will run the function (on change i think?) to format user inputted data

stevieing (approved):
It would be good to look at whether we use the store but this looks like a good implementation.

BenTopping:
> It would be good to look at whether we use the store but this looks like a good implementation.

Yeah im not sure where the best place for it to live is. It has to be manipulated in the components anyway to add custom events etc so don't think it we'd benefit much from having it in the store.

PR #1146, opened 2023-04-28
Formatter function prop in traction-input

seenanair:
- Added 'formatter' property in traction-input  - a function that determines how the value to be displayed in traction-input
- Added unit test case for formatter property
 Use case in PacbioRunWellEdit


seenanair:
@stevieing @BenTopping I have added a 'formatter' prop to traction-input and added unit test cases as well. It would be great if you can check whether it is working in the real context/ page it is displaying

stevieing (approved):

PR #1158, opened 2023-05-10
dpl-731 custom traction modal implemenation

seenanair:
Tailwind component to display an html modal dialog.


src/components/pacbio/PacbioRunPlateItem.vue
```diff
@@ -1,23 +1,43 @@
...
+      </well>
+    </Plate96SVG>
+    <WellEdit ref="modal" class="modal" :position="selectedWellPosition" @alert="alert"></WellEdit>
+  </flagged-feature>
 </template>
 
 <script>
 import Plate96SVG from '@/components/svg/Plate96SVG'
 import PlateMap from '@/config/PlateMap'
+import BWell from '@/components/pacbio/PacbioRunWellItemBS'
```
stevieing, line 26: I would maybe rename this to be PacbioRunWellItemBootstrap rather than PacbioRunWellItemBS? It just removes any ambiguity.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -165,13 +167,24 @@ export default {
...
+    async showModalForPosition(position) {
+      if (position) {
+        this.positionData = position
+      }
       // We also need to setup the well here in case state has updated since
       await this.setupWell()
-      this.$refs['well-modal'].show()
+      this.isShow = true
+      /**This need to be removed when custom_enable_modal feature flag is removed */
+      if ('b-modal' in this.$refs['well-modal'].$refs) {
```
stevieing, line 178: This looks almost the same as the hide method. Maybe it could be combined?

src/components/pacbio/PacbioSampleMetadataEdit.vue
```diff
@@ -93,9 +100,20 @@ export default {
     },
     ...mapActions(['updateRequest']),
     show() {
-      this.$refs['modal'].show()
+      this.showModal = true
+      /**This need to be removed when custom_enable_modal feature flag is removed */
+      if ('b-modal' in this.$refs['modal'].$refs) {
```
stevieing, line 105: This looks to be the same as what is going on in src/components/pacbio/PacbioRunWellEdit.vue. Maybe if we move them to the same place it will be easier to remove when the feature flag is removed?

src/components/saphyr/SaphyrEnzymeModal.vue
```diff
@@ -60,19 +63,28 @@ export default {
...
       /**
        * Hide the modal manually
        * https://vuejsdevelopers.com/2019/01/22/vue-what-is-next-tick/
        * https://bootstrap-vue.js.org/docs/components/modal/#prevent-closing
        */
+
+      /**This need to be removed when custom_enable_modal feature flag is removed */
       this.$nextTick(() => {
-        this.$refs.enzymeModal.hide()
+        if ('b-modal' in this.$refs['enzymeModal'].$refs) {
```
stevieing, line 80: Same as above ...

stevieing (changes requested):
Really good. A few minor suggestions. I still need to test it locally ...

stevieing (changes requested):
I have tried it locally and it is working everywhere apart from the pacbio library create which doesn't save.

seenanair:
> I have tried it locally and it is working everywhere apart from the pacbio library create which doesn't save.

Ohh, I will have a look into that

seenanair:
> I have tried it locally and it is working everywhere apart from the pacbio library create which doesn't save.

@stevieing It seems to work or am I doing something completely different?

https://github.com/sanger/traction-ui/assets/55585488/5011fee1-1bfd-471c-9b16-310c9ca70b51

src/components/pacbio/PacbioRunPlateItem.vue
seenanair: Done 👍

src/components/pacbio/PacbioRunWellEdit.vue
seenanair: Thought it would be better to have it separate as one is show and other hide. Also, 'showModalForPosition' is an async operation called from child component and doing some extra things _if (position) {
        this.positionData = position
      }
      // We also need to setup the well here in case state has updated since
      await this.setupWell()_
Also, thought it would be easy to support binding bootstrap events. Please share your thoughts

src/components/pacbio/PacbioSampleMetadataEdit.vue
seenanair: Done 👍

seenanair: Did you mean combining all of this similar statements _this.$refs['enzymeModal'].$refs['b-modal'].show()_ to one place. I am afraid that will require quite a bit of refactoring , with each checking it's own ref in it's own instance (this) context

src/components/PrinterModal.vue
```diff
@@ -47,28 +51,40 @@ export default {
     clearSelect() {
       this.selectedPrinterId = null
     },
+    onPrintAction() {
+      this.isShow = true
+    },
     handleOk(evt) {
       // Prevent modal from closing
-      evt.preventDefault()
+      if (evt) {
```
BenTopping, line 59: Instead of preventing the default behaviour here, vue has a neat thing where you can specify this behaviour in the component event declaration. https://vuejs.org/guide/essentials/event-handling.html#event-modifiers

src/components/shared/TractionModal.vue
```diff
@@ -0,0 +1,110 @@
...
+    /**Title to display in modal header, if don't want to use scoped slots */
+    title: {
+      type: String,
+      default: '',
+    },
+  },
+  data() {
+    return {
+      /**'ok' event emitted from footer, if there is a corresponding button in 'modal-footer' scoped slot */
+      ok: () => {
```
BenTopping, line 75: I would probably move 'ok' and 'cancel' to the methods section unless there is a specific reason they need to be under data?

BenTopping (changes requested):
Looking good, few small comments. Wondering if we can make the modal slightly more distinctive from its background? Bootstrap added a dark see-through background when the modal was open which made it easier to see against a background. Maybe we can also add this or a bigger drop shadow?

src/components/PrinterModal.vue
seenanair: Yes , .prevent actions are ver handy. But it seems like we didn't need preventDefault at all in this case with the new changes, as we are dealing with those actions within the custom modal

src/components/shared/TractionModal.vue
seenanair: Yes, you are right. It started off with a different idea and stayed there . But this can be moved to methods

seenanair:
> Looking good, few small comments. Wondering if we can make the modal slightly more distinctive from its background? Bootstrap added a dark see-through background when the modal was open which made it easier to see against a background. Maybe we can also add this or a bigger drop shadow?

Though the background drop seems to be simple, it can be tricky in cases where there are scrollable content in both the dialog and the background we are displaying. 
I have added a backdrop but it needed some  anti-patterns to deal with the scrolling. Please have a look into those

seenanair:
@BenTopping Since i have refactored traction modal with backdrop it would be great if you can check whether it is working as well when you get time

BenTopping (dismissed):
Looks good to me. Need to run prettier to fix the format CI failing

seenanair:
> Looks good to me. Need to run prettier to fix the format CI failing

Prettier is running fine locally. I don't know why it is failing in CI . Is it because of any version difference  in CI?

stevieing:
> > Looks good to me. Need to run prettier to fix the format CI failing
> 
> Prettier is running fine locally. I don't know why it is failing in CI . Is it because of any version difference in CI?

Prettier was updated a couple of weeks ago. You need to update and dependencies and install them

seenanair:
> > > Looks good to me. Need to run prettier to fix the format CI failing
> > 
> > 
> > Prettier is running fine locally. I don't know why it is failing in CI . Is it because of any version difference in CI?
> 
> Prettier was updated a couple of weeks ago. You need to update and dependencies and install them

Ohh, Ok. Thank you. I will do that

stevieing (dismissed):

PR #1159, opened 2023-05-10
BUG fix - Libraries page not showing correct source identifier

BenTopping:
Changed flattenObject to be more explicit about nested attributes. This fixes a bug where values were being overwritten incorrectly on the pacbio libraries page


seenanair (approved):
Cool!! That looks great and you have already covered it's limitation as well , so we can do any level deeper with similar keys.

PR #1161, opened 2023-05-11
DPL-725 Wildcard barcode search for the Pacbio libraries page

### LINKED ISSUE #1111: DPL-725 As a user I would like to add a wildcard barcode search to the Pacbio libraries page
**User story**
For the PacBio libraries page when searching for a barcode it should include all similar barcodes? e.g TRAC-2-35 also returns libraries like TRAC-2-351, TRAC-2-352.

Ben T, Yousra B

Yousra B

**Acceptance criteria**
- [ ] When searching by barcode on the pacbio libraries page, libraries with similar barcodes should be included in the response

**Additional context**
Design concerns:
This could become an issue down the line when there are more libraries.
E.g User is looking for TRAC-2-34 and it returns all libraries from TRAC-2-34X and TRAC-2-34XX.
Maybe look at ways we can specify whether to wildcard. Like including a * e.g TRAC-2-34* or adding a search type like barcode (loose) vs barcode (strict)

BenTopping:
Closes #1111 

* Small restyle to FilterCard
* Added ability to pass a wildcard option into FilterCard

Note:
Dependent on https://github.com/sanger/traction-service/pull/1015

### DIFF
--- a/src/components/FilterCard.vue
+++ b/src/components/FilterCard.vue
@@ -105,10 +105,10 @@ export default {
     async resetFilter() {
       this.filterValue = ''
       this.filterInput = ''
-      ;(this.filterWildcard = true),
-        await this.fetcher().then(({ success, errors }) => {
-          success ? '' : this.showAlert(errors, 'danger')
-        })
+      this.filterWildcard = true
+      await this.fetcher().then(({ success, errors }) => {
+        success ? '' : this.showAlert(errors, 'danger')
+      })
     },
   },
 }
--- a/src/components/FilterCard.vue
+++ b/src/components/FilterCard.vue
@@ -1,28 +1,37 @@
 <template>
-  <div class="flex mx-auto w-1/2 bg-gray-100 rounded-md mb-2 p-3">
-    <div class="flex w-full">
-      <div class="w-full">
+  <div class="flex mx-auto w-full max-w-[1000px] bg-gray-100 rounded-md mb-2 p-3">
+    <div class="grid grid-cols-4 gap-2">
+      <div class="flex flex-col text-left">
         <p class="flex font-semibold text-xl">Filter results</p>
         <p class="flex font-light text-gray-700 text-xs italiclist-none">
           By default returns the most recent 1000 results
         </p>
       </div>
-      <div class="flex w-full">
+      <div class="col-span-2 flex mx-auto items-center">
         <traction-input
           id="filterInput"
           v-model="filterInput"
           type="search"
           placeholder="Type to Search"
-          class="mr-5 w-full"
+          class="mr-5 w-1/2"
         />
         <traction-select
           id="filterValue"
           v-model="filterValue"
           :options="filterOptions"
-          class="w-full"
+          class="mr-5 w-1/2"
         />
+        <div v-if="wildcard" class="justify-center items-center w-1/3">
+          <label for="checkbox" class="w-1/2">Wildcard</label>
+          <input
+            id="wildcardValue"
+            v-model="filterWildcard"
+            type="checkbox"
+            class="w-1/2 bg-sbd-400"
+          />
+        </div>
       </div>
-      <div class="w-full">
+      <div class="flex items-center">
         <div class="w-full">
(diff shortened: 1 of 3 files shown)


src/components/FilterCard.vue
```diff
@@ -84,6 +105,7 @@ export default {
     async resetFilter() {
       this.filterValue = ''
```
stevieing, line 106: Maybe this should be an object instead? You have `filterValue` and `filterInput` and `filterWildcard` which would indicate to me that it should be `filter: { input: '', wildcard: '', value: '' }`. You would also then be able to make `getFilteredData` a method of the object. It would give better encapsulation

stevieing (changes requested):
I think it would be worth looking at filter as an object maybe imported from a module?? It would be easier to test and give better flexibility and reuse.

src/components/FilterCard.vue
sdjmchattie: Does that alter the efficiency of the UI?  Probably yes, but not enough to make it worth while avoiding this approach.  I don't know at what level the responsiveness of the UI operates, but if it updates UI elements because the `filter` object changed when the containing value (out of the three) hasn't changed, that could, on a larger scale, create a much larger overhead on the UI.  I doubt that's a factor here, but just wanted to identify a possible cost to what is being suggested.  It could be that Vue is far better at this than I realise.

sdjmchattie (dismissed):
Looks like there's a couple of changes from Steve you should address, but otherwise LGTM.  I don't fully understand the HTML changes, but I will leave it to you to ensure the layout works.

src/components/FilterCard.vue
BenTopping: Since getFilteredData requires 'this' and access to the fetcher im not sure it should be included as a method of filter. We could import from a filterObject from another module but this would be a bit more complicated and require splitting it up, which would also make the code more cognitively difficult to follow.

seenanair:
@BenTopping I tried a wild card  search with TRAC-2 , but it didn't show any

https://github.com/sanger/traction-ui/assets/55585488/3faf59bf-9efa-43b9-80ad-b03c74aa0cbb

BenTopping:
@seenanair There is a traction-service branch that is also required for it to work. Sorry i forgot to link it to the PR: https://github.com/sanger/traction-service/pull/1015

seenanair (approved):
Looks good 👍 
Just a minor thing, may _wildcard_ computed method can be renamed to something more intuitive (isWildCardOption) as it is looking whether the selected filter option is wild card supported, I believe

seenanair (approved):

src/components/FilterCard.vue
```diff
@@ -86,14 +86,14 @@ export default {
     }
   },
   computed: {
-    wildcard() {
+    isWildcardOption() {
```
sdjmchattie, line 89: I notice that changing this didn't require any tests to be updated with the new name.  Does this method need testing?

BenTopping: There isnt a specific test for it but it is indirectly tested here: https://github.com/sanger/traction-ui/pull/1161/files#diff-b5412883d88e30421c0da822ffc118c6e78d92461ad80bc91b19a7510785b0a6R68. I can add a test if you think one would be useful.

sdjmchattie: That's fair.  I guess the implementation details aren't important to test.  Only that the user selects an option to be a wildcard filter and that is the one that gets marked as such.  I would recommend updating that test though to make each of the three filters the wildcard option.  You can use `it.each` (https://vitest.dev/api/#test-each) to run the test multiple times with the value of the filter to enable.  i.e. `[1, 2, 3]`.  Just testing the first one means the coverage isn't optimal.  Glad that there is a test though!

BenTopping: yeah that sounds good, I have added something along those lines.

sdjmchattie (approved):

stevieing (approved):
I think we could also do something slightly differe

PR #1172, opened 2023-05-19
Dpl 766 custom alert message 

seenanair:
Traction message refactoring to allow html message display using tailwind styling
 - enable_custom_alert feature flag to flip between bootstrap( blert) and html alert message
 - compatibility with all existing bootstrap features supported


src/components/TractionMessage.vue
```diff
@@ -22,23 +49,47 @@ export default {
...
     },
+    /**
+     * Returns tailwind colors based on 'type' property with default being grey.
+     *  The supported values for 'type' property are as follows
+     * 'success' -  will be displayed in green color
+     * 'error'  -  will be displayed in red color
+     */
+    color() {
+      let value = { message: 'bg-gray-300 text-gray-400', icon: '' }
+      if (this.type === 'success') {
```
BenTopping, line 82: Might be nice to have all the types as an object or something where the keys are the type. Then the computed can just search the object and return the value or fallback onto the default.

BenTopping (approved):
Looks good, very seamless from the old alert. One small optional comment. Also Id be inclined to make the text slightly darker / more distinct from the background. But I have colour seeing issues anyway so probably worth a second opinion 😆

src/components/TractionMessage.vue
seenanair: Sure, this seems like it will make it more confined. will try 👍

seenanair:
> Looks good, very seamless from the old alert. One small optional comment. Also Id be inclined to make the text slightly darker / more distinct from the background. But I have colour seeing issues anyway so probably worth a second opinion 😆 

Thank you , Ben. I will try with a darker version, we are inclusive :)

stevieing (dismissed):
Looks good to me. Checked it locally and all working.

seenanair:
> Looks good, very seamless from the old alert. One small optional comment. Also Id be inclined to make the text slightly darker / more distinct from the background. But I have colour seeing issues anyway so probably worth a second opinion 😆
Done. Hope that's better now

stevieing (approved):
All working as expected :)

stevieing (approved):

PR #1178, opened 2023-05-22
Window jumping issue fix in traction modal

seenanair:


BenTopping (commented):
Looks good. A few unrelated things I found. Line 119 isn't needed as the visible watcher will run resetScrollbar itself when display value changes, so currently it is getting run twice on modal close. Also found another small bug. When using the bootstrap modal the scrolling doenst get reset on modal close. Can send an example if need be.

stevieing (approved):

PR #1189, opened 2023-06-02
1117 dpl 728 custom component to replace b form file

enegoktan:
Closes #

*
*
* ...


src/components/ont/OntPoolEdit.vue
```diff
@@ -154,15 +163,16 @@ export default {
         this.busy = false
       })
     },
+
     // Allows users to upload a file to autopopulate the pool's selected libraries
-    async uploadFile(newFile) {
-      if (newFile === null) {
+    async uploadFile(evt) {
+      if (evt === null || evt.target.files === null || evt.target.files.length == 0) {
```
stevieing, line 169: Advisory:This could also be `if (!!evt?.target?.files?.length)` but that reverses the result so not sure if it works here.

src/components/pacbio/PacbioPoolEdit.vue
```diff
@@ -146,13 +153,17 @@ export default {
         this.busy = false
       })
     },
-    async uploadFile(newFile) {
-      if (newFile === null) {
+
+    // Allows users to upload a file to autopopulate the pool's selected libraries
+    async uploadFile(evt) {
+      if (evt === null || evt.target.files === null || evt.target.files.length == 0) {
```
stevieing, line 159: Advisory:This could also be`if (!!evt?.target?.files?.length)` but that reverses the result so not sure if it works here.

stevieing (approved):
Well done for getting this finished in your last week! As it is used in multiple places it could be turned into a component but it looks like there are significant differences between each use case for that to be tricky so I would vote to leave it as is and maybe try if we have to do this again.

stevieing (approved):
Nice. Feel free to merge and deploy to UAT.

PR #1201, opened 2023-06-08
DPL-786 Qc result upload fix

### LINKED ISSUE #1198: DPL-786 [BUG] - On the qc results upload page the file outline is always green and re-enable does not work correctly
**Describe the bug**
On the qc results upload page the file outline is always green and re-enable does not work correctly on invalid upload the line around the file input remains green (as it is supposed to be green when the input is valid).

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page qc results upload page
2. Upload file 
3. Click on button Upload file
4. See message QC results successfully uploaded
5. Click re-enable
6. See message QC results successfully uploaded

**Expected behaviour**
The page should reset and Uplaod file should be re-enabled

**Screenshots**
Screenshot 2023-06-07 at 14 57 29
.

* Windows or Mac?

**Additional context**

harrietc52:
Closes #1198

* When clicking “Upload File”, and it fails, have the “Choose File” input box be outlined in red (not green)
* When Re-enable is clicked, after either a successful or unsuccessful upload, reset the “Choose File” input to empty, and remove any outline colour.


src/components/qcResultsUpload/QcResultsUploadForm.vue
```diff
@@ -71,18 +72,27 @@ export default {
...
       disableUpload: null,
+      uploadSuccessful: null,
     }
   },
   computed: {
     qcResultUploadsRequest: ({ api }) => api.traction.qc_results_uploads.create,
     border() {
-      if (this.file === null) return 'border-0'
-      else {
+      if (this.uploadSuccessful === true) {
```
seenanair, line 81: May be It would be better to have a comment here to avoid ambiguity in future that  describes the expected styles associated with three states - true, false and null

src/components/qcResultsUpload/QcResultsUploadForm.vue
```diff
@@ -71,18 +72,27 @@ export default {
       usedBySelected: 'extraction',
       busy: null,
       disableUpload: null,
+      uploadSuccessful: null,
```
seenanair, line 75: just a minor thing, May be It would be better to have a comment here to avoid ambiguity in future that  describes the expected  states - true, false and null

seenanair (approved):
Looking good 👍

src/components/qcResultsUpload/QcResultsUploadForm.vue
stevieing, line 81: you could just have this.uploadSuccessful as the true is implicit. You could also use Also use the ternary operator i.e. `return this.uploadSuccessful ? 'rounded border border-green-500' : 'rounded border border-red-500'` as anything that is not true should be false e.g. null.

stevieing (dismissed):
Just one advisory comment. Didn't check whether it is working locally.

src/components/qcResultsUpload/QcResultsUploadForm.vue
harrietc52: Yes I was wondering if there was a cleaner way!
The only thing I can think, is that here false = red, and null = no colour. So they have different behaviour, so maybe we need them all?

seenanair: I initially thought the same that it can be replaced with a conditional operator, but then saw it can represent three different states null, true and false not null = false in this case. That;s why I thought a comment would be better ,so it will be clear.

src/components/qcResultsUpload/QcResultsUploadForm.vue
harrietc52: Thank you, I've added a comment

src/components/qcResultsUpload/QcResultsUploadForm.vue
harrietc52: I've added a comment but will have a play to see if it can be refactored. Thanks both

seenanair: I think if we enable --strictNullChecks in typescript config file (that we usually do) , it will complain and we need to explicitly use boolean|null as the type for the variable.
But , the checking will work I think because we are  specifically checking it against false (this.uploadSuccesful == false) and not using if(!uploadSuccesful)

harrietc52: Its cool that you have the typescript knowledge to predict this in advance, to help with the transition over later! 👏

seenanair: Not an expert, just enough to keep up the work. More to explore

harrietc52: @stevieing @seenanair So is the current implementation of this okay do we think? If so, I'll deploy to UAT today

seenanair: It looks fine to me 👍

seenanair (approved):

harrietc52:
**Deployment**

`ansible-playbook -vD deploy_traction_ui.yml -i environments/uat -e "github_branch=v6.13.0-5242861358-develop reload_nginx=true "`

harrietc52:
Develop -> Master PR [here](https://github.com/sanger/traction-ui/pull/1205)

src/components/qcResultsUpload/QcResultsUploadForm.vue
stevieing: Happy for it go but just worth discussing at some point if it is worth using a type. We use it effectively in lighthouse ui https://github.com/sanger/lighthouse-ui/blob/develop/components/StatusAlert.vue. In this case we are using null true and false to indicate 3 different states so to remove ambiguity we could use an enum which would give a more explicit idea of what is going on. Maybe something worth doing if and when we attempt to move to typescript ...

seenanair: Also, I have come across a similar use in ONTFlowCell with three states. So, if it is a type maybe it can re-used

PR #1220, opened 2023-06-19
Add labware handler component

BenTopping:
Closes #

* Added labwareMap component to build generic labware
* Replace PacbioRunWellItem and PacbioRunPlateItem with new components using labwareMap


stevieing (changes requested):
All looks fine as is. Just one thing:
-When you drag a tube onto a well it would give improved feedback if the well was highlighted as it hovers over.

stevieing (approved):

src/components/labware/LabwareMap.vue
```diff
@@ -0,0 +1,63 @@
...
+      required: true,
+      default: 0,
+    },
+  },
+  methods: {
+    createPosition(rowNumber, columnNumber) {
+      if (rowNumber > 26) {
+        return `${rowNumber},${columnNumber}`
+      }
+      const aCharCode = 'A'.charCodeAt(0)
```
harrietc52, line 57: What's this for?

harrietc52 (commented):
All looks good, definitely an improvement. 

Not sure whether is it worth talking to Seena as she might be clearer on ways to further abstract. 

As I'm not sure if this is possible but I was imagining something like : instead of hardcoding 8 and 12 for rows and columns in PacbioRunPlate, it would be something more removed. e.g. 
PacbioRunPlate would be called something more generic, like 96Plate. 

Then PacbioRunShow would state that it requires a 96Plate, which would be an const of some sort, which maps to config for 96Plate as key, with the row/column numbers as values. 

That enum (96Plate) would be passed into the LabwareMap, which builds it based on the config it pulls. The general Labware builder would still have a slot for the passed in "Well" component

Again, not sure if this over simplifies it, but might be worth pondering on / asked Seena what her thoughts are.

seenanair:
> All looks good, definitely an improvement.
> 
> Not sure whether is it worth talking to Seena as she might be clearer on ways to further abstract.
> 
> As I'm not sure if this is possible but I was imagining something like : instead of hardcoding 8 and 12 for rows and columns in PacbioRunPlate, it would be something more removed. e.g. PacbioRunPlate would be called something more generic, like 96Plate.
> 
> Then PacbioRunShow would state that it requires a 96Plate, which would be an const of some sort, which maps to config for 96Plate as key, with the row/column numbers as values.
> 
> That enum (96Plate) would be passed into the LabwareMap, which builds it based on the config it pulls. The general Labware builder would still have a slot for the passed in "Well" component
> 
> Again, not sure if this over simplifies it, but might be worth pondering on / asked Seena what her thoughts are.

Yes, that's definitely a very abstract way of doing it which gives flexibility in wiring different labware types also (if needed) very easily by extending  labware configuration settings file. I didn't mention it at this time because I was not sure how useful it is in Traction context as opposed to STAN where we support many different labware types. One advantage is then we don't have any labware specific configuration embedded in any of the

BenTopping:
@harrietc52 @seenanair Yep definitely an approach we can use and I think it will be worth implementing it in the future if we end up with more labware types. In this case the PacbioRunPlate has some specific behaviour like managing the WellEdit modal, but this could be abstracting out to the run page in the future if we want to re-purpose the 96plate for other things. If you are both happy then I think ill keep the current implementation as is, but maybe add a todo as a future note in case we run into this again?

harrietc52 (approved):
Awesome, nice one 😄

PR #1242, opened 2023-06-30
Dpl 791 styling changes

seenanair:
1. ONT Run page reformatting

- Create,reset,back buttons to display at top
- Spacing between headers
- Vertical align of instrument and state fields

2. Subtle border between Filter panel and table to give visual clarity in all pages with filter option

3. 'Generate sampleshheet' changed to button with a download icon in ONTRunIndex and PacbioRunIndex
  - Added Download icon

4. Pool/new page styling changes for Pacbio and ONT

5. ONTFlowCell changes
  -Change the tick colour in the TractionFieldError component to match the darker green colour in the Flowcell component
- Change bootstrap colours to tailwind border-warning border-danger etc

6. Sorting button fix in Firefox

### DIFF
--- a/src/views/ont/ONTPoolIndex.vue
+++ b/src/views/ont/ONTPoolIndex.vue
@@ -1,7 +1,6 @@
 <template>
   <DataFetcher :fetcher="fetchOntPools">
     <FilterCard :fetcher="fetchOntPools" :filter-options="filterOptions" />
-    <div class="flex mt-2 mb-3 border" />
     <div class="flex flex-col">
       <div class="clearfix">
         <traction-pagination
--- a/src/views/ont/ONTPoolIndex.vue
+++ b/src/views/ont/ONTPoolIndex.vue
@@ -1,6 +1,7 @@
 <template>
   <DataFetcher :fetcher="fetchOntPools">
     <FilterCard :fetcher="fetchOntPools" :filter-options="filterOptions" />
+    <div class="flex mt-2 mb-3 border" />
     <div class="flex flex-col">
       <div class="clearfix">
         <traction-pagination
--- a/src/views/ont/ONTRun.vue
+++ b/src/views/ont/ONTRun.vue
@@ -24,9 +24,7 @@
       >
     </div>
     <ONTRunInformation></ONTRunInformation>
-    <div class="mt-4">
-      <ONTRunInstrumentFlowcells></ONTRunInstrumentFlowcells>
-    </div>
+    <ONTRunInstrumentFlowcells></ONTRunInstrumentFlowcells>
   </DataFetcher>
 </template>
 <script>
--- a/src/views/ont/ONTRun.vue
+++ b/src/views/ont/ONTRun.vue
@@ -1,29 +1,32 @@
 <template>
   <DataFetcher :fetcher="provider">
-    <router-link :to="{ name: 'ONTRunIndex' }">
-      <traction-button id="backToRunsButton" class="float-right">Back</traction-button>
-    </router-link>
-    <traction-button
-      v-if="newRecord"
-      id="resetButton"
-      type="reset"
-      theme="default"
-      class="float-right"
-      @click="newRun"
-      >Reset</traction-button
-    >
+    <div class="flex flex-row justify-end mt-2">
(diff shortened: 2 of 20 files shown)


stevieing (changes requested):
All working as expected. Just a couple of things:
- I wonder if the download sample sheet button could be a different colour? It is currently the same as the edit button
- On the pool new page there is a slight misalignment on the vertical for 1a and 1b

I think ONT run could do with a bit of a redesign but we agreed to roll that out into another story.

src/views/ont/ONTPoolIndex.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <DataFetcher :fetcher="fetchOntPools">
     <FilterCard :fetcher="fetchOntPools" :filter-options="filterOptions" />
+    <div class="flex mt-2 mb-3 border" />
```
BenTopping, line 4: Instead of adding this to all pages could we move it into the FilterCard component as a wrapper?

src/views/ont/ONTRun.vue
```diff
@@ -1,29 +1,32 @@
...
+        class="float-right"
+        :theme="currentAction.theme"
+        :disabled="!runValid"
+        @click="runAction"
+        >{{ currentAction.label }}</traction-button
+      >
+    </div>
     <ONTRunInformation></ONTRunInformation>
-    <ONTRunInstrumentFlowcells></ONTRunInstrumentFlowcells>
+    <div class="mt-4">
```
BenTopping, line 27: Probably move this styling into the ONTRunInstrumentFlowcells component like its done in ONTRunInformation

src/views/pacbio/PacbioRunIndex.vue
```diff
@@ -81,7 +82,9 @@
             :href="generateSampleSheetPath(row.item.id)"
             class="text-primary"
           >
-            Generate Sample Sheet
+            <traction-button :id="generateId('generate-sample-sheet', row.item.id)">
+              Generate Sample Sheet <DownloadIcon class="pl-1" />
```
BenTopping, line 86: DownloadIcon is a nice touch 😄

src/views/ont/ONTPoolIndex.vue
seenanair: Yes, makes sense in the way  how it is used currently and it is required in all  calling contexts. I didn't do that because having a split-bar below doesn't actually can be considered as this component styling but rather belongs to the choice at parent level. 
 I will move it to FilterCard and probably make it as an customisable style based on a prop - does that sound good?

BenTopping: Yeah I guess it does sit in a bit of a grey area. Yes I think adding custom class props or something would be a good solution.

seenanair:
> All working as expected. Just a couple of things:
> 
> * I wonder if the download sample sheet button could be a different colour? It is currently the same as the edit button
> * On the pool new page there is a slight misalignment on the vertical for 1a and 1b
> 
> I think ONT run could do with a bit of a redesign but we agreed to roll that out into another story.

I have given a different colour to download initially, but with all that different colors on different buttons (Start etc) it looks like a rainbow effect with having too many colours  :) , that's why changed to this.  But, I will give it a go with another color  and see how it goes

seenanair:
> All working as expected. Just a couple of things:
> 
> * I wonder if the download sample sheet button could be a different colour? It is currently the same as the edit button
> * On the pool new page there is a slight misalignment on the vertical for 1a and 1b
> 
> I think ONT run could do with a bit of a redesign but we agreed to roll that out into another story.
Done

Is this color ok?
Screenshot 2023-07-03 at 12 20 06

New story added for ONTFlowCell restyling
https://github.com/sanger/traction-ui/issues/1244

src/views/ont/ONTRun.vue
seenanair: Done

src/views/ont/ONTPoolIndex.vue
seenanair: Done

stevieing (approved):

PR #1259, opened 2023-07-12
DPL-805 UAT feedback

### LINKED ISSUE #1239: DPL-805 Display serial number on run show page for revio plates
**User story**
As a user (Craig) I would like the serial number to be displayed on the run show page for revio plates. The serial number is a subsection of the sequencing plate barcode. See attached images.

The scanned barcode for the Revio sequencing plate is 28 characters:
e.g.
1021188000301570037320231019
 
The serial number is characters 16 to 20:
e.g.
102118800030157**00373**20231019

Craig C

Craig C

**Acceptance criteria**
- [x] when i edit or create a sequencing run for revio the serial number should be displayed

**Additional context**

2plate_traction1

harrietc52:
Closes #1239

* Add feedback to SKBB
* Make Serial number stand out more

### DIFF
--- a/src/components/pacbio/PacbioRunPlate.vue
+++ b/src/components/pacbio/PacbioRunPlate.vue
@@ -4,16 +4,26 @@
       <div v-for="plateNumber in labware.plateCount" :key="plateNumber" class="w-full">
         <div class="text-left mx-5 mb-5 flex flex-col">
           <traction-label classes="text-left my-2">Plate number: {{ plateNumber }}</traction-label>
-          <traction-label classes="text-left">Sequencing Kit Box Barcode:</traction-label>
-          <traction-input
-            id="sequencing-kit-box-barcode"
-            v-model="runItem.plates[plateNumber].sequencing_kit_box_barcode"
-            :value="runItem.plates[plateNumber].sequencing_kit_box_barcode"
-            placeholder="Sequencing Kit Box Barcode"
-            type="text"
-            :data-attribute="`sequencing_kit_box_barcode-${plateNumber}`"
-          />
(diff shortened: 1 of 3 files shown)


src/components/pacbio/PacbioRunPlate.vue
```diff
@@ -82,9 +92,26 @@ export default {
     onWellClick(position, plateNumber) {
       this.$refs.modal.showModalForPositionAndPlate(position, plateNumber)
     },
+    sequencingKitBoxBarcode(plateNumber) {
+      return this.runItem.plates[plateNumber].sequencing_kit_box_barcode
+    },
     // Only used for Revio runs
     serialNumber(plateNumber) {
-      return this.runItem.plates[plateNumber].sequencing_kit_box_barcode.substring(15, 20)
+      return this.sequencingKitBoxBarcode(plateNumber).substring(15, 20)
```
stevieing, line 100: Advisory: I am wondering if it would be worth moving the validation to the labware object itself? Happy to think about that as part of my refactor so as not to delay.

stevieing (approved):
Just wondering whether we could add a valid method to the PacbioRunSystems? That way you wouldn't need any logic in the component.

src/components/pacbio/PacbioRunPlate.vue
harrietc52: Sounds sensible, but ye maybe worth doing once we have refactored to have the PlateList and Plate components, as sequenceingKitBoxBarcode may then be a data attribute and serialNumber a computer etc.

stevieing: Indeed. I will add it to the list.

PR #1268, opened 2023-07-20
DPL-800 - Vue3 upgrade

BenTopping:
Closes #1227 

Changes proposed in this pull request:

* Updated Vue dependency and changes to support breaking changes
* Updated Vuex dependency
* Updated VueRouter dependency
* Update Vue test utils dependency to support vue 3 changes
* Update Vite support for vue3
* Update node to v18.17.0
* Removed VueSelecto in favour of Vue3Selecto
* Remove vue-simple-spinner in favour of custom component

Issue #1240, opened 2023-06-30
DPL-746-4 - As developers we would like to refactor the Pacbio Runs page to make it more flexible and maintainable


stevieing:
**User story**
As developers we would like to refactor the Pacbio Runs page to make it more flexible and maintainable

**Who are the primary contacts for this story**
Steve / Harry / Stephen / Ben

**Who is the nominated tester for UAT**
PSD

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] Create a Lucid Chart diagram to show how instrument type should work
- [x] Move instrument type to store
- [x] Create instrument type in PacbioRunInfoEdit based on system name
- [x] Modify PacbioPlateList to use instrumentType from store
- [x] Create a Lucid Chart to show how run components can interact with the store
- [x] Create a method to extract well data by plate number and position
- [x] Move pacbio run plates to pacbio run plate item
- [x] Create getOrCreatePlate action
- [x] Modify pacbioRunPlateItem to use getOrCreatePlate
- [x] merge develop with serial number changes
- [x] move serial number validation to object
- [x] modify current methods in run to use new structure
- [x] modify current actions in runCreate to use new structure
- [x] modify current mutations in runCreate to use new structure
- [x] move instrument type functionality to PacbioInstrumentTypes
- [x] update process diagrams to reflect latest changes
- [ ] Create a wireframe to merge the 2 well components together and remove the modal (delayed need input from @seenanair)

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

Issue #1227, opened 2023-06-23
DPL-800 - As developers we would like to upgrade Traction from Vue 2.7 to 3 as Vue 2 is due to be deprecated in December 20203


stevieing:
**Description**
As developers we would like to upgrade Traction from Vue 2.7 to 3 as Vue 2 is due to be deprecated in December 20203
Work (updated as identified)
- [x] Change app creation in main.js
- [x] Change Vue.set and Vue.delete to just use standard value set and delete
- [x] Remove vitals plugin-vue2 and reference in vite config
- [ ] Identify breaking features
- https://v3-migration.vuejs.org/breaking-changes/v-model.html
- https://v3-migration.vuejs.org/breaking-changes/emits-option.html
- [x] Review custom shared components
- [x] Fix UI missing styling
- [x] Ensure CI is operating correctly
- [ ] Ensure deployment builds are working

**VueX** (https://vuex.vuejs.org/guide/migrating-to-4-0-from-3-x.html#installation-process) 
- [x] Change ./store/index.js to use createStore directive
- [x] Change app creation to register store from ./store/index.js

**VueRouter** (https://router.vuejs.org/guide/) & (https://router.vuejs.org/guide/migration/)
- [x] Change router.js to use createRouter directive
- [x] Change app creation to register router from ./router

**VueSWRV** (https://docs-swrv.netlify.app/features.html)
- [x] Remove ./lib/composition_wrapper
- [x] Update site config to remove reference to composition wrapper
- [x] Update useSWRV to take a ref value as URL

**Vue test utils**
- [x] Update syntax to use latest vue test utils features
- [x] Update setup files to use global configs
- [x] Ensure all tests are passing

**Who the primary contacts are for this work**
Steve / Seena / Ben

**Knowledge or Stake holders**
Other people that may have specific knowledge about this work or have a stake in how it is implemented. e.g. John Smith is an expert on x

**Additional context or information**
This story originally focused on just researching the upgrade, however upon investigation it made more sense to research and implement at the same time to prevent duplicated effort.

stevieing:
Chat GPT answer:

If you're using Vite instead of Vue CLI, the process of upgrading your Vue 2.7 application to Vue 3 will have some differences. Here's a modified guide specifically for upgrading with Vite:

[...]

PR #1284, opened 2023-08-01
Develop into Master

StephenHulme:

- #1246
  - #1240
- #1269
- #1270 
- #1268
  - #1227
- #1271
- #1273
- #1274


BenTopping:
Think this one deserves a major release version bump!

BenTopping:
> > Think this one deserves a major release version bump!
> 
> According to SemVer a major bump is for a backward incompatiable change to the public API (the users of the service). A minor bump may be used if substantial new functionality or improvements are introduced within the private code.
> 
> I'd suggest that these are improvements within the private code that are not visible to the users - but I'm very happy to be swayed!

SemVer for front-end web interfaces is tricky as they don't have an API, so if we followed it exactly we would never have major changes. In my experience we normally add major changes for big user-side changes or for major developer-side changes. It may be worth looking at previous Major versions in the GitHub releases to see if this one would classify as one.

That being said I understand your argument and strict SemVer wouldn't require a major change. So I would be happy to go with whatever you decide.

BenTopping (approved):

PR #1286, opened 2023-08-02
BUG fix - Removed unsupported __vueParentComponent references

BenTopping:

- Removed references to __vueParentComponent as they are removed in production env 

Quite an interesting one. __ vue __ was removed in vue3 but we used this to access an elements attributes during runtime for vueSelecto elements. This was replaced during the vue3 upgrade with __vueParentComponent but this value is removed in production builds. So it worked locally but not when compiled.
https://stackoverflow.com/questions/64119722/what-is-the-vue-3-equivalent-of-vue-2s-vue

The fix in this case is to use the javascript `getAttribute` method as ID is exposed as an attribute. 
https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute

### DIFF
--- a/src/components/ont/OntPlateSelectedList.vue
+++ b/src/components/ont/OntPlateSelectedList.vue
@@ -110,10 +110,10 @@ export default {
     },
     onSelect(e) {
       e.added.forEach((el) => {
-        this.selectWellRequests(el.__vueParentComponent.attrs.id)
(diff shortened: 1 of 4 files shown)


BenTopping:
> Much cleaner! Is this the fix for the RT ticket from earlier today?

Yes. It was definitely a bit hacky before so nice to have removed it.

PR #1302, opened 2023-08-11
Dpl 792 colour handling

### LINKED ISSUE #1210: DPL-792 Refactor Traction UI colour handling
**Describe the Housekeeping**
As a developer, we would like the Traction UI to have a single source of truth of its styling colours used throughout the application. Currently, many different components specify the colours of successful/ error etc, and some don't match. We would like to refactor out of the components where the colour codes of specified, so they are all the same. This came out of: https://github.com/sanger/traction-ui/issues/1112#issuecomment-1589014291 

**Blocking issues**
Describe any other issues or tickets that may be blocking this change.

**Additional context**

This could be done as part of [DPL-791](https://github.com/sanger/traction-ui/issues/1203), or by itself

seenanair:
Closes #1210 

- Success, failure and warning colour definitions in tailwind.config.cjs file
- 'tailwind.css' file creation in src/styles
- Defined custom styles in newly created 'tailwind.css
- Refactoring of all references to new colour definitions
- Test case updates

### DIFF
--- a/src/components/pacbio/PacbioPoolTubeItem.vue
+++ b/src/components/pacbio/PacbioPoolTubeItem.vue
@@ -152,7 +152,7 @@ export default {
       // when the pool is invalid we have a red background with red text
       // TODO: modify red background for hover to be slightly softer. Need new tailwind colour
       return `m-1 border-2 cursor-pointer rounded-md pb-2  ${
-        this.valid ? 'hover:bg-gray-200' : 'bg-red-200 hover:bg-red-300 text-red-800'
+        this.valid ? 'hover:bg-gray-200' : 'failure-style'
       }`
     },
   },
--- a/src/components/pacbio/PacbioPoolTubeItem.vue
+++ b/src/components/pacbio/PacbioPoolTubeItem.vue
@@ -150,7 +150,6 @@ export default {
     classObject() {
       // when the pool is valid we have a gray background
       // when the pool is invalid we have a red background with red text
-      // TODO: modify red background for hover to be slightly softer. Need new tailwind colour
       return `m-1 border-2 cursor-pointer rounded-md pb-2  ${
         this.valid ? 'hover:bg-gray-200' : 'failure-style'
       }`
--- a/src/styles/tailwind.css
+++ b/src/styles/tailwind.css
@@ -2,16 +2,16 @@
 @import '../../node_modules/@sanger/ui-styling/tailwind.css';
 
 /* Define your custom styles */
-.success-style {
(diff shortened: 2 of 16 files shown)


src/components/pacbio/PacbioPoolTubeItem.vue
```diff
@@ -152,7 +152,7 @@ export default {
       // when the pool is invalid we have a red background with red text
       // TODO: modify red background for hover to be slightly softer. Need new tailwind colour
```
BenTopping, line 153: Can remove this TODO now?

src/styles/tailwind.css
```diff
@@ -0,0 +1,17 @@
+/* Import Tailwind's base styles from ui-styling package. Optional if you don't want to use them.**/
+@import '../../node_modules/@sanger/ui-styling/tailwind.css';
+
+/* Define your custom styles */
+.success-style {
```
BenTopping, line 5: is the general practice to add `-style` to the end?

BenTopping: Also are these needed if we only use them in 1 or two places?

BenTopping (commented):
I think this approach works, but im wondering whether it makes more sense to add them to the 'ui-styling' project? Similar to  how 'Sanger Dark Blue' works https://github.com/sanger/ui-styling/blob/develop/tailwind.config.js#L128

src/components/pacbio/PacbioPoolTubeItem.vue
seenanair: Done

src/styles/tailwind.css
seenanair: i have just kept two , in case it can be reused later. Changed names to 'success-icon', 'success-message','failure-icon', 'failure-message'

seenanair:
> I think this approach works, but im wondering whether it makes more sense to add them to the 'ui-styling' project? Similar to how 'Sanger Dark Blue' works https://github.com/sanger/ui-styling/blob/develop/tailwind.config.js#L128

Added 'success', 'warning' and 'failure' colours to ui-styling package

seenanair:
> I think this approach works, but im wondering whether it makes more sense to add them to the 'ui-styling' project? Similar to how 'Sanger Dark Blue' works https://github.com/sanger/ui-styling/blob/develop/tailwind.config.js#L128

Added 'success', 'warning' and 'failure' colours to ui-styling package

BenTopping (approved):
Looks good 👍

PR #1313, opened 2023-08-25
Condense PacBio Runs Table and Show Versions

StephenHulme:
#### Changes proposed in this pull request:

- Show SMRT-Link version numbers
- Condense the width of the runs table
  - by hiding disabled buttons
  - displaying multiple plates above one another
  - with small stylistic changes

See #1334 for a visual representation of all the badges.

Before:

Before

After:

After

If smrt-link versions don't load:
Screenshot 2023-09-11 at 10 12 06

### DIFF
--- a/tests/unit/views/pacbio/PacbioRunIndex.spec.js
+++ b/tests/unit/views/pacbio/PacbioRunIndex.spec.js
@@ -1,6 +1,6 @@
 import PacbioRunIndex from '@/views/pacbio/PacbioRunIndex'
 import Response from '@/api/Response'
-import { mount, store, Data, flushPromises } from '@support/testHelper'
+import { mount, store, Data, flushPromises, nextTick } from '@support/testHelper'
 
 describe('PacbioRunIndex.vue', () => {
   let wrapper, pacbioRunIndex, mockRuns, mockVersions
@@ -39,14 +39,6 @@ describe('PacbioRunIndex.vue', () => {
     })
   })
 
-  // must run before the version badge tests below (I can't get the mocks to apply independently)
-  describe('smrt-link versions missing', () => {
-    it('displays an error badge if the version is not found', () => {
-      const badgesMissing = wrapper.find('tbody').findAll('.badge')
-      expect(badgesMissing[0].text()).toEqual('< ! >')
-    })
-  })
-
   describe('version badge', () => {
     let badges
 
@@ -72,6 +64,14 @@ describe('PacbioRunIndex.vue', () => {
         const version_name = version.name.split('_')[0] // keep only the version number, dropping everything after the underscore
         expect(badge.text()).toEqual(version_name)
       })
+    })
+
+    it('displays an error badge if the version is not found', async () => {
+      // Remove smrtLinkVersions from state
+      store.state.traction.pacbio.runCreate.resources.smrtLinkVersions = {}
+      await nextTick()
+      const badgesMissing = wrapper.find('tbody').findAll('.badge')
+      expect(badgesMissing[0].text()).toEqual('< ! >')
     })
   })
 


BenTopping (changes requested):
Nice improvement! Looks a lot cleaner. Couple comments about use of styles mainly.

src/components/shared/TractionBadge.vue
```diff
@@ -0,0 +1,66 @@
...
+      default: 'blue',
+      required: false,
+    },
+  },
+}
+</script>
+<style scoped>
+/* badges */
+.badge {
+  @apply inline-flex items-center rounded-lg px-1 py-0 my-1 text-xs font-medium;
```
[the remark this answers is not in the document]
seenanair: I agree , it is better to avoid mixing Tailwind CSS with inline styles if possible to help maintainability and to maintain consistency across project.

src/components/shared/table/TractionTable.vue
```diff
@@ -381,3 +381,9 @@ export default {
   },
 }
 </script>
+<style scoped>
```
seenanair, line 384: same as above - it would be better to avoid <style>

seenanair (commented):
Looking good with a new component to display badge  👍

src/views/pacbio/PacbioRunIndex.vue
```diff
@@ -185,7 +185,8 @@ export default {
       return import.meta.env.VITE_TRACTION_BASE_URL + '/v1/pacbio/runs/' + id + '/sample_sheet'
     },
     generateVersionColour(versionIndex) {
-      const colours = ['gray', 'red', 'yellow', 'green', 'blue', 'indigo', 'purple', 'pink']
+      // colours used in TractionBadge
+      const colours = ['red', 'orange', 'yellow', 'green', 'blue', 'indigo', 'purple', 'pink']
```
seenanair, line 189: Just a thought - maybe, we can move these colours to TractionBadge class itself and can be exported from there, just in case this need to be accessed in a similar way in other places in future

seenanair (dismissed):
Looking good 👍

src/views/pacbio/PacbioRunIndex.vue
```diff
@@ -171,8 +199,14 @@ export default {
     redirectToRun(runId) {
       this.$router.push({ path: `/pacbio/run/${runId || 'new'}` })
     },
-
     ...mapActions('traction/pacbio/runs', ['fetchPacbioRuns', 'updateRun']),
+    ...mapActions('traction/pacbio/runCreate', ['fetchSmrtLinkVersions']),
+    async provider() {
```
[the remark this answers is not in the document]
BenTopping: Nice. I think this logic should live in the DataFetcher and it gets passed the fetchers object, but for now this is good as otherwise we would have to change all the other places that DataFetcher is used, maybe add a TODO or something.

I also noticed that the FilterCard uses this function so it will refetch the SMRTLink versions each time which isn't necessary but its not a big issue and i'm not sure if there is a nice way to avoid it.

tests/unit/views/pacbio/PacbioRunIndex.spec.js
```diff
@@ -1,17 +1,16 @@
...
-describe('Runs.vue', () => {
-  let wrapper, runs, mockRuns
+describe('PacbioRunIndex.vue', () => {
+  let wrapper, pacbioRunIndex, mockRuns, mockVersions
 
   beforeEach(async () => {
     mockRuns = new Response(Data.PacbioRuns).deserialize.runs
-    const get = vi.spyOn(store.state.api.traction.pacbio.runs, 'get')
-    get.mockReturnValue(Data.PacbioRuns)
+    vi.spyOn(store.state.api.traction.pacbio.runs, 'get').mockResolvedValue(Data.PacbioRuns)
```
BenTopping, line 10: Does the SMRTLink Version call need to be mocked here as well or is it not called?

src/components/shared/TractionBadge.vue
```diff
@@ -0,0 +1,89 @@
...
+      default: 'sanger-dark-blue',
+      required: false,
+      validator: within(...Object.keys(flatColourClasses)),
+    },
+  },
+  computed: {
+    colourStyle: ({ colour }) => flatColourClasses[colour],
+  },
+  // colour lists like {warm: ['red', 'orange', 'yellow'], cool: ['blue', 'indigo', 'violet']}
+  colours: Object.keys(colourClasses).reduce((acc, curr) => {
```
BenTopping, line 85: Pretty neat. Didn't know you could add class methods to vue components like this.

BenTopping (approved):

PR #1338, opened 2023-09-15
pacbio/tubes migration from VueX to Pinia

seenanair:
- Defined `stores` folder for Pinia
- Pinia store defined for pacbio/tubes - src/stores/pacbioTubes.js added
   - Actions, getters method refactored for Pinia
   - Mutations removed
   - Unit test cases added
- VueX store for pacbio/tubes removed - src/store/traction/pacbio/tubes
    - Unit test cases removed


seenanair:
@BenTopping @stevieing I couldn't find any mutations/getters for this Vuex store being called from anywhere else. Please correct if otherwise and I will make the required changes.

src/stores/pacbioTubes.js
```diff
@@ -0,0 +1,50 @@
...
+  state: () => {
+    return {
+      tractionTubes: [],
+    }
+  },
+  getters: {
+    //Pinia_migration_todo: This is migrated from the VueX store, but it needs to be changed to  Pinia store once the VueX store is removed
+    tubeRequest: () => store.state.api.traction.pacbio.tubes,
+    //Pinia_migration_todo: This is migrated from the VueX store, but it needs to be changed to  Pinia store once the VueX store is removed
+    requestsRequest: () => store.state.traction.pacbio.requests,
```
BenTopping, line 15: should this be `store.state.api.traction.pacbio.requests`

src/stores/pacbioTubes.js
```diff
@@ -0,0 +1,50 @@
...
+      if (!barcode) {
+        return false
+      }
+      const libraryTube = await this.getTubeForBarcode(barcode)
+      return this.validateLibraryTube(libraryTube)
+    },
+    async getTubeForBarcode(barcode) {
+      const request = this.tubeRequest
+      const promise = request.get({ filter: { barcode: barcode }, include: 'pools' })
+      const response = await handlePromise(promise)
```
BenTopping, line 29: We are trying to move away from `handlePromise`. We now use `handleResponse` from src/api/ResponseHelper. Do you think you could change this function to use handleResponse instead?

src/stores/pacbioTubes.js
```diff
@@ -0,0 +1,50 @@
...
+        return response.deserialize.tubes[0]
+      }
+    },
+    validForRunCreation(pool) {
+      return (
+        pool.volume &&
+        pool.concentration &&
+        pool.template_prep_kit_box_barcode &&
+        pool.insert_size &&
+        true
```
BenTopping, line 40: what does the `&& true` do here?

tests/unit/stores/pacbioTubes.spec.js
```diff
@@ -0,0 +1,117 @@
...
+  //Test the isLibraryBarcodeValid action
+  it('should return false if there is no pool', async () => {
+    const store = usePacbioTubesStore()
+    const sampleTube = new Response(Data.TractionTubesWithPacbioPools).deserialize.tubes[0]
+    store.getTubeForBarcode = vi.fn().mockReturnValue(sampleTube)
+    const result = await store.isLibraryBarcodeValid('TRAC-1')
+    expect(result).toEqual(false)
+  })
+
+  //Test the isLibraryBarcodeValid action
```
BenTopping, line 39: Group tests by putting them inside a separate 'describe' block instead of adding the comments at the top of each. Each `describe('isLibraryBarcodeValid', () => { <tests> })`

src/store/traction/pacbio/index.js
```diff
@@ -1,4 +1,4 @@
-import pacbioTubesModule from '@/store/traction/pacbio/tubes'
+//import pacbioTubesModule from '@/store/traction/pacbio/tubes'
```
stevieing, line 1: can this be removed?

src/stores/pacbioTubes.js
```diff
@@ -0,0 +1,50 @@
...
+
+export const usePacbioTubesStore = defineStore('pacbioTubes', {
+  state: () => {
+    return {
+      tractionTubes: [],
+    }
+  },
+  getters: {
+    //Pinia_migration_todo: This is migrated from the VueX store, but it needs to be changed to  Pinia store once the VueX store is removed
+    tubeRequest: () => store.state.api.traction.pacbio.tubes,
```
BenTopping, line 13: We can manually make the API object here instead of using the one built from the VueX store so we dont tie any dependencies.

src/stores/pacbioTubes.js
```diff
@@ -0,0 +1,50 @@
...
+import handlePromise from '@/api/PromiseHelper'
+import store from '@/store'
+
+export const usePacbioTubesStore = defineStore('pacbioTubes', {
+  state: () => {
+    return {
+      tractionTubes: [],
+    }
+  },
+  getters: {
```
stevieing, line 11: Do we need these at all? I'm just wondering if we could just use the store directly. I am not sure how many places these are used.

BenTopping (dismissed):
Couple comments. I am not sure where we actually use the PacBio tubesStore. Is it still needed?

src/stores/pacbioTubes.js
stevieing, line 29: It might be better to use handleResponse rather than handlePromise. handlePromise is deprecated and handleResponse also offers better error handling.

src/stores/pacbioTubes.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+      const libraryTube = await this.getTubeForBarcode(barcode)
+      return this.validateLibraryTube(libraryTube)
+    },
+    async getTubeForBarcode(barcode) {
+      const request = this.tubeRequest
+      const promise = request.get({ filter: { barcode: barcode }, include: 'pools' })
+      const response = await handlePromise(promise)
+      if (response.successful && !response.empty) {
+        return response.deserialize.tubes[0]
```
stevieing, line 31: I think we could move away from deserialize here but it might be better left for later.

stevieing (changes requested):
Looking goos especially as there is now less code. There are lots of things that could be improved but they are not within the scope of this story so just suggested things that are easy to change.

seenanair:
@BenTopping @stevieing Thank you for the comments. I will work on this.

stevieing (approved):

PR #1345, opened 2023-09-22
dpl-907 migrating pacbio runs store to Pinia

seenanair:
- Create Pinia store for pacbio runs -  /stores/pacbioRuns
- Updated test cases
- Removed VueX store 
- Removed test cases for Vuex store

@stevieing @BenTopping 
Areas that might require refactoring while migrating the store code is commented with a //TODO tag


tests/unit/stores/pacbioRuns.spec.js
```diff
@@ -0,0 +1,102 @@
+import { setActivePinia, createPinia } from 'pinia'
```
stevieing, line 1: Would it be worth adding this to the testHelper and importing? The same as we have done with Vuex.

tests/unit/views/pacbio/PacbioRunIndex.spec.js
```diff
@@ -1,15 +1,41 @@
 import PacbioRunIndex from '@/views/pacbio/PacbioRunIndex'
 import Response from '@/api/Response'
 import { mount, store, Data, flushPromises, nextTick } from '@support/testHelper'
+import { createTestingPinia } from '@pinia/testing'
```
stevieing, line 4: Just trying to get my head round the differences between creating an active pinia and using the testing Pinia so not entirely sure what is going on here?

tests/unit/views/pacbio/PacbioRunIndex.spec.js
```diff
@@ -1,15 +1,41 @@
...
+import { usePacbioRunsStore } from '../../../../src/stores/pacbioRuns'
```
stevieing, line 5: You can just use 

`import { usePacbioRunsStore } from '@/src/stores/pacbioRuns'. `

Same in a few other places.

tests/unit/views/pacbio/PacbioRunIndex.spec.js
```diff
@@ -1,15 +1,41 @@
...
 describe('PacbioRunIndex.vue', () => {
   let wrapper, pacbioRunIndex, mockRuns, mockVersions
 
   beforeEach(async () => {
     mockRuns = new Response(Data.PacbioRuns).deserialize.runs
-    vi.spyOn(store.state.api.traction.pacbio.runs, 'get').mockResolvedValue(Data.PacbioRuns)
-
-    wrapper = mount(PacbioRunIndex, { store })
+    const spy = vi.fn().mockResolvedValue({ success: true, data: Data.PacbioRuns.data })
+    const { wrapperObj } = factory({
```
stevieing, line 25: Again this looks quite clever but I am not entirely sure what is going on.

stevieing (changes requested):
Seems good. Not entirely sure what is going on in a few places so maybe worth some comments.

tests/unit/views/pacbio/PacbioRunIndex.spec.js
seenanair: `createPinia` is for creating a Pinia instance to test pinia stores and `createTestingPinia` is to mock Pinia store to test a component that uses Pinia store. 
This is how it is recommended in Pinia documentation https://pinia.vuejs.org/cookbook/testing.html

tests/unit/views/pacbio/PacbioRunIndex.spec.js
seenanair: The factory is a builder function, which allows to create a mock instance of pinia,  with the given options that specifies how the store mockup should behave and what state it should be in while instantiating the component. This particular example doesn't make use of the function properly because we are using the same settings for all the test cases, but it can be useful in cases where we need to mock it differently for different test cases. 
Hope that make sense.

stevieing: Sure is it worth putting this as some documentation so that when we all start using Pinia we can follow your lead?

seenanair: Sure, will do :) 👍

stevieing (approved):

PR #1352, opened 2023-10-02
fix(pacbio-runs-view): fix filter card not responding to filters

BenTopping:
This bug was caused by the provider method not taking parameters which apply to the pacbio runs request. There is a wider issue that FilterCard and DataFetcher need to handle multiple requests more elegantly. I will create an issue for this.

This fix works because we can assume smrtLinkVersions are already populated by the page load so we can just search pacbioRuns directly instead of handling multiple requests and re-fetching smrtlinkversions.


BenTopping:
> Woops, I think I caused this! 🙈 Thanks for catching and fixing 🙏

No worries! These things happen, especially when there are abstracted methods like this. It also highlights a hole in our testing as it should have been caught by a test.

PR #1355, opened 2023-10-02
DPL-871 View old runs

### LINKED ISSUE #1300: DPL-871 -  As a developer I would like the ability to view old runs in Traction
**User story**
As a developer I would like to add a fall back option for SMRTLink Versions and System Names so that old PacBio runs in Traction can still be viewed. Currently if you go to Traction production and try view an old run (TRACTION-RUN1), it errors when you try and view the run.

Ben T

Ben T

**Acceptance criteria**
- [ ] When I go to a run with no SMRTLink Version or System Name I can still see the run plate and wells

KatyTaylor:
Fixes around the 'system name' and 'smrt link version' fields
- System name wasn't showing because Sequel I and Sequel II weren't present in the config, due to being old systems.
- Therefore the instrument type data wasn't getting populated and things were erroring that tried to access it
- I've just copied the Sequel IIe config to create entries for the old systems - will need to find out the correct values here
- This allows the Plate section to render, and the value to appear in the 'system name' drop down.
- smrt link version wasn't showing because the database entry said 'inactive', and the API only retrieved 'active' values
- Tweaked the API to allow through active values (in traction service), and filtered them out from the drop down unless the record value was an inactive one

See comment https://github.com/sanger/traction-ui/issues/1300#issuecomment-1748946591 and screenshots on story https://github.com/sanger/traction-ui/issues/1300#issuecomment-1749062904

Closes #1300 

- 

### DIFF
--- a/src/components/pacbio/PacbioRunInfoEdit.vue
+++ b/src/components/pacbio/PacbioRunInfoEdit.vue
@@ -87,19 +87,14 @@ export default {
     ...mapGetters(['runItem', 'smrtLinkVersionList', 'smrtLinkVersion', 'instrumentType']),
     // Makes an array of objects with value and text properties to make
     // the options of smrt-link-version select drop-down list.
+    // Only includes 'active' versions in the list, unless this record already has an inactive version as its value.
     smrtLinkVersionSelectOptions() {
-      var activeVersionsList = []
-      for (const versionKey in this.smrtLinkVersionList) {
-        const version = this.smrtLinkVersionList[versionKey]
-        // Only includes 'active' versions in the list, unless this record already has an inactive version as its value
-        if (version.active == true || version.id == this.smrtLinkVersion.id) {
-          activeVersionsList.push({
-            value: version.id,
-            text: version.name,
-          })
-        }
-      }
-      return activeVersionsList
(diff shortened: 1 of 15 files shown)


src/components/pacbio/PacbioRunInfoEdit.vue
```diff
@@ -85,13 +85,21 @@ export default {
...
+    // Makes an array of objects with value and text properties to make
+    // the options of smrt-link-version select drop-down list.
     smrtLinkVersionSelectOptions() {
-      // Returns an array of objects with value and text properties to make
-      // the options of smrt-link-version select drop-down list.
-      return Object.values(this.smrtLinkVersionList).map(({ id, name }) => ({
-        value: id,
-        text: name,
-      }))
+      var activeVersionsList = []
```
stevieing, line 91: would it be possible just to use.
    ```
  return Object.values(this.smrtLinkVersionList).map(({ id, name }) => ({
        value: id,
        text: name,
      }).filter( version => version.active)

```
 I am not sure just a thought?

KatyTaylor: Yes probably, I thought there would be a better way! Thanks

stevieing (approved):

PR #1360, opened 2023-10-05
DPL-888 - Page by page pagination

### LINKED ISSUE #1317: DPL-888 As a developer I would like to implement page by page pagination
**User story**
As a developer I would like to implement page by page pagination. This involves editing the get requests for the paginated pages in traction to only pull back the data for the current page. This should take into account the page number and number of records per page. Adding this will **significantly** improve the speed of traction when there are a large number of records to display. This can be seen in production where index pages take a long time to load.

Ben T

N/A

**Acceptance criteria**
- [ ] Ont and Pacbio Index pages only fetch the data related to the current page
- [ ] Page changes on index pages fetch the correct requests. e.g. page 2 size 20 should fetch the 21-41st records

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1317 

Requires traction-service branch: DPL-888-page-by-page-pagination

- Adds pagination to pacbio index pages
- Changes traction-pagination and filter-card to utilise query params in the URL
- Adds filter card to ont runs page


BenTopping:
**Current status (06/10/23)**

- Pacbio Index pages have been reworked to use URL query pagination.
- Saphyr and ONT index pages still need to be reviewed. Do we want to update saphyr? Do we need to add filters to ONT run (would be an improvement)
- Feature flags. We need to think about how we are going to feature flag this. Should we duplicate pagination components (traction-pagination, filter-card) to keep old functionality and add the ability to switch between the two?
- QueryParamsHelper Mixin. Is there a more Vue3'y way we can do this? Instead of using a mixin could we use a useQueryParams.js file where we import the methods into each file that needs it?
- FilterCard watchers. Currently filter card watches for updates in page_size and page_number so it knows whether to refetch data. Should this be done here or somewhere else, e.g. could we do this in traction-pagination where we initally change the variables.
- TableHelper. Can we get rid of this TableHelper mixin as it is used to support front-end pagination which is no longer required.

src/api/createRequest.js
```diff
@@ -64,11 +64,12 @@ const parametersToString = (attributes, parameter = undefined) => {
  * @param {Object} filter - query filters
  * @param String include - query include
  * @param {Object} fields - query fields
+ * @param {Object} page - query page
  * @return String
  * Turns a list of parameters into a query string
  */
-const buildQuery = ({ filter = {}, include = '', fields = {} } = {}) => {
-  const queryString = parametersToString({ filter, include, fields })
+const buildQuery = ({ page = {}, filter = {}, include = '', fields = {} } = {}) => {
```
stevieing, line 71: Advisory:  wonder if it would be worth turning this into a type? I thought about this when I did it originally but needed to stop somewhere. get, find and create all use the build query method. Seems like it might be an idea to pass that type

src/router.js
```diff
@@ -25,6 +25,11 @@ import ONTRunIndex from '@/views/ont/ONTRunIndex'
 import ONTRun from '@/views/ont/ONTRun'
 import ONTSampleIndex from '@/views/ont/ONTSampleIndex'
 
+function checkPaginationParams(to) {
```
stevieing, line 28: What is this for? Might be worth adding some docs.

src/store/traction/ont/pools/actions.js
```diff
@@ -289,18 +289,18 @@ export default {
...
    */
-  fetchOntRequests: async ({ commit, rootState }, filter) => {
+  fetchOntRequests: async ({ commit, rootState }, filter, page) => {
     const request = rootState.api.traction.ont.requests
-    const promise = request.get({ filter: filter })
+    const promise = request.get({ page, filter })
     const response = await handleResponse(promise)
 
-    const { success, data: { data } = {}, errors = [] } = response
+    const { success, data: { data, meta = {} } = {}, errors = [] } = response
```
stevieing, line 297: Again we are making lots of the same changes across the board. Seems like it points to a code smell. Should we use a shared method or a type?

src/views/PipelineView.vue
```diff
@@ -44,6 +44,10 @@ export default {
      * is handled by the router.push method**/
     setSource(indx) {
```
stevieing, line 45: why not use index?

src/views/ont/ONTRunIndex.vue
```diff
@@ -104,6 +106,15 @@ export default {
       this.$router.push({ path: `/ont/run/${runId || 'new'}` })
     },
     ...mapActions(['fetchOntRuns']),
+    async fetchRuns() {
+      const page = { size: this.page_size.toString(), number: this.page_number.toString() }
```
stevieing, line 110: Again in a lot of places we are doing exactly the same thing. Should page be an object we can import and standardise across the app.

stevieing (changes requested):
Great piece of work. I have tested it locally and it is working correctly.
A lot of the comments I have added are advisory but it would be worth considering abstracting some bits out so it is easier to reuse fpr future work. I didn't comment in all of the places I saw these issues. 
Also worth adding some documentation in some places.

src/views/ont/ONTRunIndex.vue
BenTopping: Agreed, I have moved this behaviour into the queryParamsHelper, it also means we don't need to import lots of single use data in the setup.

src/store/traction/ont/pools/actions.js
BenTopping: Yeah agreed. Should be a separate story I think as all API calls could be refactored like this.

src/store/traction/ont/actions.js
```diff
@@ -8,21 +8,21 @@ export default {
    * @param rootState the vuex rootState object. Provides access to the current state
    * @param commit the vuex commit object. Provides access to mutations
    */
-  fetchOntRuns: async ({ commit, rootState }) => {
+  fetchOntRuns: async ({ commit, rootState }, filter, page) => {
     const request = rootState.api.traction.ont.runs
-    const promise = request.get({ include: 'instrument' })
+    const promise = request.get({ page, filter, include: 'instrument' })
```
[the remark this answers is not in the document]
BenTopping: Agreed. Again makes sense in a separate story as it affects all the requests.

src/api/createRequest.js
BenTopping: Yep. Job for a separate story, maybe something like "Add consistent typing across API requests builders?"

src/components/FilterCard.vue
```diff
@@ -83,42 +79,48 @@ export default {
...
     },
   },
   methods: {
     async getFilteredData() {
-      let searchValue = this.filter.input
-      if (this.isWildcardOption && this.filter.wildcard) {
-        // If wildcard is selected, add it to the search string
-        searchValue += ',wildcard'
+      if (this.page_number == 1) {
+        await this.getData()
```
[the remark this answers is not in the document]
BenTopping: Yes its a good point. Couple things to bear in mind as to why I have done it this way:

- We don't want to refetch every time a character is typed into the filter as that will cause a lot of unnecessary requests.
- The filter needs to be reset to page: 1 when its searched as the new results may have less pages that the existing results which would break the page until reset.

That said I will take another look and see what I can do as I think you are right that there is probably a nicer way to encapsulate all the data requests. Open to suggestions.

src/components/FilterCard.vue
```diff
@@ -83,42 +79,48 @@ export default {
...
+      } else {
+        // This triggers the page_number watcher and ensures we are on the first page
+        this.page_number = 1
```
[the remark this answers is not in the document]
BenTopping: I have renamed this.

src/views/PipelineView.vue
BenTopping: 🤷‍♂️ Will update it.

src/components/FilterCard.vue
```diff
@@ -83,42 +79,48 @@ export default {
...
+      page_number and page_size:
+      We ensure the page is paginated before fetching data.
+      This catches an edge case where the watchers are still active when navigating
+      between paginated and unpaginated routes
+    */
+    page_size() {
+      this.$route.meta?.paginated ? this.getData() : ''
+    },
+    page_number() {
+      this.$route.meta?.paginated ? this.getData() : ''
```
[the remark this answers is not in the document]
seenanair: Ok, I got it. I think, in that way it make more sense for the watchers to be here rather than in pagination. Again, as you mentioned it doesn't bring any specific advantage by moving this fetching to one generic component to another.

BenTopping:
@seenanair I had a go at trying to move the watchers to the useQueryParams function and passing in the fetcher in the useQueryParams definition during the setup function. However there are issues with access to the vuex methods during setup because setup doesn't have access to 'this'. I think we can do what you are suggesting when we move to Pinia but until then I think we will have to stick with the current setup. i will add a comment to the helper file saying this.

src/components/FilterCard.vue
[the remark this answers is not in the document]
seenanair: Yep, I agree, those points are very valid. 
I don't whether it make sense, but to have a  state inside composable that depends on query params , which watches over values to do fetching.

seenanair:
> @seenanair I had a go at trying to move the watchers to the useQueryParams function and passing in the fetcher in the useQueryParams definition during the setup function. However there are issues with access to the vuex methods during setup because setup doesn't have access to 'this'. I think we can do what you are suggesting when we move to Pinia but until then I think we will have to stick with the current setup. i will add a comment to the helper file saying this.

Sorry, I just something very similar to this idea without seeing your message. I totally agree and it can very well wait until it state management changes to Pinia. 👍

seenanair (approved):

seenanair (commented):
Yep, that description is very clear to me. 👍

seenanair (dismissed):

stevieing (approved):

seenanair (dismissed):

PR #1375, opened 2023-10-17
Added types for select options in PacbioRunInfoEdit.

stevieing:
Closes #

- Modify select options in PacbioRunInfoEdit to use a type. This would potentially allow it to be used elsewhere.

### DIFF
--- a/src/lib/SelectOptionsTypes.js
+++ b/src/lib/SelectOptionsTypes.js
@@ -20,7 +20,7 @@ const SelectOptionsType = ({ key, name, list }) => {
    * @param {object} currentItem - the current item to compare to
    * @returns {array} - a list of items that are active or have the same key as the current item
    */
-  const filter = (currentItem) => {
+  const filterByActiveOrKey = (currentItem) => {
     return list.filter((item) => item.active || item[key] === currentItem[key])
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/SelectOptionsTypes.js
```diff
@@ -0,0 +1,52 @@
...
+    value: key,
+    text: name,
+  }
+
+  /*
+   * filter returns a list of items that are active or have the same key as the current item
+   * @param {object} currentItem - the current item to compare to
+   * @returns {array} - a list of items that are active or have the same key as the current item
+   */
+  const filter = (currentItem) => {
```
BenTopping, line 23: I would change the name of this filter to something else as 'filter' is a JS keyword and it isn't descriptive of what it is doing. Alternatively you could make this function more agnostic to take the filters like 'active' and 'key' as parameters and call the function filterMatch or something.

BenTopping (changes requested):
Looks like a nice refactor. Just one comment about the filter function.

src/lib/SelectOptionsTypes.js
stevieing: I did think about making it more flexible but would wait until we have further selects that could do with the same treatment. filter is not a JS keyword as far as I know. It is a method on Array which is why it would not be a problem here as we have created our own type.

BenTopping: Yeah thats sensible. Yeah sorry thats what I meant, that its an existing JS array method name, its not a problem as you said, but your change makes it more readable.

BenTopping (approved):

PR #1401, opened 2023-11-09
DNA control complex box barcode added for v12 Sequel IIe

seenanair:
dna_control_complex_box_barcode field enabled  for Sequel IIe

### DIFF
--- a/src/components/pacbio/PacbioRunInfoEdit.vue
+++ b/src/components/pacbio/PacbioRunInfoEdit.vue
@@ -16,7 +16,7 @@
       </traction-field-group>
 
       <traction-field-group
-        v-if="!smrtLinkVersionv12"
+        v-if="!isRevio"
         label="DNA Control Complex Box Barcode"
         for="dna-control-complex-box-barcode"
       >
@@ -109,6 +109,9 @@ export default {
     },
     smrtLinkVersionv12() {
       return /^v12/.test(this.smrtLinkVersion.name)
+    },
+    isRevio() {
+      return this.instrumentType.key == PacbioInstrumentTypes.Revio.key
     },
   },
   methods: {
--- a/src/components/pacbio/PacbioRunInfoEdit.vue
+++ b/src/components/pacbio/PacbioRunInfoEdit.vue
@@ -107,11 +107,8 @@ export default {
         this.instrumentType,
       )
     },
-    smrtLinkVersionv12() {
-      return /^v12/.test(this.smrtLinkVersion.name)
-    },
     isRevio() {
-      return this.instrumentType.key == PacbioInstrumentTypes.Revio.key
+      return this.instrumentType.key === PacbioInstrumentTypes.Revio.key
     },
   },
   methods: {
--- a/src/stores/utilities/run.js
+++ b/src/stores/utilities/run.js
@@ -1,5 +1,3 @@
-import { PacbioInstrumentTypes } from '@/lib/PacbioInstrumentTypes'
-
 /**
  *
  * @param {String} plateNumber The number of the plate e.g. 1
@@ -159,7 +157,7 @@ const createPayload = ({ id, run, plates, wells, smrtLinkVersion, instrumentType
       type: 'runs',
       id,
       attributes: {
-        ...createRunPayload(run, instrumentType),
+        ...run,
         pacbio_smrt_link_version_id: smrtLinkVersion.id,
         system_name: instrumentType.name,
         plates_attributes: Object.values(plates)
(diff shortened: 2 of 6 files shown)


src/components/pacbio/PacbioRunInfoEdit.vue
```diff
@@ -110,6 +110,9 @@ export default {
     smrtLinkVersionv12() {
       return /^v12/.test(this.smrtLinkVersion.name)
     },
+    isRevio() {
```
stevieing, line 113: Probably best to do === adds an extra level of checking.

src/components/pacbio/PacbioRunInfoEdit.vue
```diff
@@ -110,6 +110,9 @@ export default {
     smrtLinkVersionv12() {
```
stevieing, line 110: No longer needed?

src/stores/utilities/run.js
```diff
@@ -188,16 +188,13 @@ const hasPlateAttributes = ({ sequencing_kit_box_barcode, wells_attributes }) =>
  * @param {smrtLinkVersion} - The SMRT Link Version of the run
  * @returns {run} - A run object
  */
-const SMRT_LINK_VERSION_V12_SEQUEL_IIE = 'v12_sequel_iie'
 const SMRT_LINK_VERSION_V12_REVIO = 'v12_revio'
```
stevieing, line 191: Is this only used in 1 place. Sometimes wonder if we need a constant for single usage ...

stevieing (changes requested):
Just a couple of minor things. If you get those done I will approve so you can get it onto UAT tomorrow. I am not in but Ben should be able to help if you need anything. Maybe better to deploy from branch?

src/components/pacbio/PacbioRunInfoEdit.vue
seenanair: Ok 👍

src/components/pacbio/PacbioRunInfoEdit.vue
seenanair: Removed

src/stores/utilities/run.js
seenanair: Yes, it's only used in one place.  It's right, no need to define a constant if only used in one place( Perhaps it makes it more clear than directly used in function). I will remove it

seenanair: Also, with only one string it can better be changed to if (smrtLinkVersion.name === 'v12_revio')

seenanair: One other question, I think it makes more sense to pass instrumentType as we are  linking dna_control_barcode field to to the system rather than the SMRT Link Version  and it matches the comparison on UI side. What do you think?

stevieing: Makes sense.

src/stores/utilities/run.js
```diff
@@ -188,16 +190,11 @@ const hasPlateAttributes = ({ sequencing_kit_box_barcode, wells_attributes }) =>
...
-  // dna_control_complex_box_barcode is currently an optional field in the service
-  // as it is not required for v12 Sequel IIe, but it for the other versions
-  if (
-    [SMRT_LINK_VERSION_V12_SEQUEL_IIE, SMRT_LINK_VERSION_V12_REVIO].includes(smrtLinkVersion.name)
-  ) {
+
+const createRunPayload = (run, instrumentType) => {
+  // If the system is Revio, remove redundant dna_control_complex_box_barcode
+  // as it is not required for Revio, but required for the other systems
+  if (instrumentType.key === PacbioInstrumentTypes.Revio.key) {
```
stevieing, line 197: Do we need to do this? If it is a Revio it is already null

stevieing (commented):
Just 1 minor thing to check.

src/stores/utilities/run.js
seenanair: Yes. but, I thought this method is to make sure that  no unintended attributes are set for a smrtLinkVersion which was the case before, now changed to instrumentType. If we edit a SequleII run and changed the instrumentType to Revio

stevieing: Good thought. If you move it to the setInstrumentData method that would probably be better. That is where we remove all of the data if the instrument type is changed. https://github.com/sanger/traction-ui/blob/develop/src/stores/pacbioRunCreate.js#L432C29-L432C29

seenanair: That sounds a better place. I will check it out 👍  Thank you

stevieing (approved):

PR #1412, opened 2023-11-17
DPL-981 store migration from VueX to Pina for ont/runs

seenanair:
- Created ontRoot, ontRuns store
- Updated stores in components
- Unit test cases added
- Removed Vue stores


src/components/ont/runs/ONTFlowcell.vue
```diff
@@ -131,80 +133,81 @@ export default {
     },
     flowcellId: {
       get() {
-        const flowcell = this.currentRun().flowcell_attributes.find(
-          (flowcell) => flowcell.position == this.position,
-        )
-        if (flowcell) {
-          return flowcell.flowcell_id
+        if (this.currentRun) {
```
stevieing, line 136: Not entirely clear what has changed here. 

A way to remove all of the logic would be to create a null object for current run which would return everything you need when currentRun is undefined.

src/components/ont/runs/ONTFlowcell.vue
```diff
@@ -131,80 +133,81 @@ export default {
...
-       * white - if both flowcellId and barcode fields are empty
-       * yellow - if one of flowcellId and barcode fields are valid and other is empty
-       */
-      flowcell_bg_colour() {
-        if (!this.isFlowcellIdExists && !this.isBarcodeExists)
-          return 'border border-3 border-gray-300'
+    poolTubeBarcode() {
+      //This is keep poolTubeBarcode in sync with the Pinia store state  (option api way)
+      const ontRunStore = useOntRunsStore()
+      const flowcell = ontRunStore.currentRun.flowcell_attributes.find(
```
stevieing, line 153: same here. If you set currentRun in the store to be something like:

`{  flowcell_attributes: [ { position: 1, tube_barcode: '' }] }`

or you could return a blan flowcell. It is essentially adding a type so you don't need to do lot's of checking.

src/components/ont/runs/ONTFlowcell.vue
```diff
@@ -131,80 +133,81 @@ export default {
...
+        (flowcell) => flowcell.position == this.position,
+      )
+      if (flowcell) {
+        return flowcell.tube_barcode
+      }
+      return ''
+    },
+    /**Displays green if valid, red if invalid and no border if empty */
+    flowcell_id_field_colour() {
+      return this.isFlowcellIdExists
```
stevieing, line 163: Same here. If you make flowcell a type you can delgate the checks to the type itself which means things become more encapsulated. At the moment there are quite a few methods in flowcell which are checking the attributes of flowcell.

src/stores/ontRoot.js
```diff
@@ -0,0 +1,101 @@
...
+    resources: {
+      // The main source of run information. runs are indexed by id.
+      runs: {},
+      // The main source of instrument information. instrument are indexed by id.
+      instruments: {},
+    },
+  }),
+  getters: {
+    runs: (state) => {
+      try {
```
stevieing, line 32: Same here. Not entirely sure what is happening so worth adding some documentation and again thinking about using types so something is always returned.

src/stores/ontRuns.js
```diff
@@ -0,0 +1,171 @@
...
+      const run = this.currentRun
+      const request = this.runRequest
+
+      const existingInstruments = ontRootStore.instruments
+      const instrument_id = existingInstruments.find((i) => i.name == run.instrument_name).id
+
+      //TODO: This need to be refactored to use the Pinia once ont/ppols is migrated
+      const existingPools = store.getters['traction/ont/pools/pools']
+
+      const flowcell_attributes = run.flowcell_attributes
```
stevieing, line 42: it looks like createRun and updateRun are also using complex methods to create stuff and doing the same thing which could be resolved by using types.m I am also not sure what fc is.

stevieing (commented):
Maybe worth going through this and deciding whether we do any of my suggestions.

It is clear that some of this could do with a refactor but it might be worth rolling into another story. One of the things I have noticed is that the ONT runs page performs poorly and this might be part of the reason.

Nothing to do with what you have done! We made the decision at the time to do enough to get it working.

stevieing (changes requested):
This is really good. I think there is a lot more we could do with respect to the flowcells e.g. add the colour stuff to the flowcell object itself but I think we can save that for another day.

The only thing I am going to ask you to do is add more documentation the objects and methods. I have referred to a file as an example https://github.com/sanger/traction-ui/blob/develop/src/api/JsonApi.js. It would be great if we could have the params and returnType where possible. This will make it more maintainable.

seenanair:
> This is really good. I think there is a lot more we could do with respect to the flowcells e.g. add the colour stuff to the flowcell object itself but I think we can save that for another day.
> 
> The only thing I am going to ask you to do is add more documentation the objects and methods. I have referred to a file as an example https://github.com/sanger/traction-ui/blob/develop/src/api/JsonApi.js. It would be great if we could have the params and returnType where possible. This will make it more maintainable.

Thank you @stevieing .
As you mentioned, I thought about moving colours to here, but then I felt it is more conceptually right to allow the component to define it. We can definitely think about it, if we go for another round of optimisation.

For the documentation, I hope you meant utilities/flowCell.js file. I will have a look into the file and add more documentation. 👍

stevieing (approved):

seenanair:
@stevieing One request, would you please test the ont run and the 'new run' page manually as well when you get time?

stevieing (approved):
It might be worth comparing the utilities for ont and pacbio runs. They are essentially doing a lot of the same stuff but slightly differently. Again we can look at this again when we come to refactor ont runs due to the performance issues.

PR #1453, opened 2023-12-14
DPL 950 - NT barcodes import

seenanair:
- GeneralReception page update
    - New source option "Sequencescape Tubes"
    - Print support
         - Displays only when source  selected is "Sequencescape Tubes"
         - Displays human barcodes when user enters a valid barcode
         - Printer option selection
         - Support priniting on 'Print' button click
    - Fetching labware data from Sequencescape on entering/scanning barcodes
    - Import functaionlity modification
        
- SequencescapeTubes added in receptions
- sequencescapeUils.js added - all generic functions used between receptions/Sequencescape and receptions/SequenscapeTubes
- Unit and e2e test case updates
- Resolving import paths


src/lib/receptions/Sequencescape.js
```diff
@@ -0,0 +1,142 @@
...
+    labwareTypes,
+    labwareRequestConfig,
+  })
+}
+
+/**
+ *
+ * @returns {Array} Array of attribute keys
+ */
+const getAttributeKeys = () => {
```
stevieing, line 130: I think you can turn this into a 1 liner
`return Object.keys(labwareTypes).flatMap((type) => labwareTypes[type].attributes)`

src/lib/receptions/SequencescapeTubes.js
```diff
@@ -0,0 +1,99 @@
...
+    included,
+  })
+
+  return {
+    barcode: labware.attributes.labware_barcode.human_barcode,
+    // build the request and sample objects
+    ...buildRequestAndSample({ aliquot, study, sample, sample_metadata, requestOptions }),
+  }
+}
+const labwareTypes = {
```
stevieing, line 55: labwareTypes seem to be the same in both receptions. Maybe we can move it to the utils?

src/lib/receptions/index.js
```diff
@@ -25,30 +22,30 @@ const ReceptionTypes = {
...
     value: 'SamplesExtraction',
   },
+  SequencescapeTubes: {
+    name: 'sequencescape-tubes',
+    text: 'Sequencescape Tubes',
+    value: 'SequencescapeTubes',
+  },
 }
 
 // TO MODIFY - remove versions once dpl_877_reception_request is enabled by default
```
stevieing, line 32: I think you can remove this comment now you have removed the flag?

src/lib/receptions/sequencescapeUtils.js
```diff
@@ -80,93 +50,38 @@ const labwareForReception = async ({ requests, barcodes, requestOptions }) => {
...
-    // build the request and sample objects
-    ...buildRequestAndSample({ aliquot, study, sample, sample_metadata, requestOptions }),
+  if (success) {
+    return { data, included }
+  } else {
+    throw new Error(`Problem talking to Sequencescape: ${errors}`)
   }
 }
 
-const labwareTypes = {
```
stevieing, line 148: As per my previous comment I think we could put this back to remove duplication in the receptions.

src/views/GeneralReception.vue
```diff
@@ -223,9 +265,16 @@ export default {
     requestOptions: defaultRequestOptions(),
     barcodes: '',
     modalState: defaultModal(),
+    // labwareData is used to store the input barcodes as well as the data fetched from sequencescape which is then imported into traction
+    labwareData: {
+      inputBarcodes: [], // inputBarcodes is used to store the barcodes that the user has inputted
+      foundBarcodes: new Set(), // foundBarcodes is used to store the barcodes that have been found in sequencescape
```
stevieing, line 271: Nice. Using sets is a great way to prevent duplicates.

src/views/GeneralReception.vue
```diff
@@ -235,96 +284,163 @@ export default {
...
     },
     clearModal() {
       this.modalState = defaultModal()
     },
+
     showModal(message) {
       this.modalState = { visible: true, message }
     },
-    importFailed({ message }) {
+    //Handler for deleting barcodes from the input field
```
stevieing, line 311: Not entirey sure what is going on here but you may be able to turn it into a map.

stevieing (changes requested):
A few things to look at but looking good.

src/lib/receptions/SequencescapeTubes.js
seenanair: Actually they are different in both -
For Sequencescape it is, 
`const labwareTypes = {
  plates: {
    type: 'plates',
    attributes: 'plates_attributes',
    transformFunction: transformPlate,
  },
  tubes: {
    type: 'tubes',
    attributes: 'tubes_attributes',
    transformFunction: transformTube,
  },
}
`
For SequenceTubes , it is
`const labwareTypes = {
  plates: {
    type: 'plates',
    attributes: 'plates_attributes',
    transformFunction: transformPlate,
  },
  tubes: {
    type: 'tubes',
    attributes: 'tubes_attributes',
    transformFunction: transformTube,
  },
}
`

Probably , we can have the default in utils as 
`const labwareTypes = {
  tubes: {
    type: 'tubes',
    attributes: 'tubes_attributes',
    transformFunction: transformTube,
  },
}
`
and in Sequencescape , we can modify this generic tye to append the 'plates' as well. But , I am not sure whether it improves much as there are only two of them. Please comment.

src/views/GeneralReception.vue
seenanair: This handler is designed for an edge case that addresses inconsistencies arising when the user enters barcodes and subsequently deletes them using the backspace or delete key. When the user inputs data, we fetch that information upon pressing the enter key, and our `labwareData` object keeps the necessary details from sequencescape fetch for printing or importing. 
However, if the user removes the barcode by pressing delete or backspace, the data stored in our state may become inconsistent with what is present in the input barcode field. Therefore, it is essential to delete all information corresponding to the deleted input barcode. Failing to do so could result in printing or importing information related to barcodes that are no longer in the input barcode field.  
Hope it make sense.

As we are keeping three different array fields  in `labwareData` for inputBarcodes, foundBarcdes(from SS) and attributes (again from SS for import) I thought forEach is better to make changes for each of them. 

Please let me know your thoughts on this.

seenanair: The alternative solution to completely eliminate this issue and this handler is to inform the user that, after making any changes in input barcode field( whether it is delete or any edit), they should press 'Enter'. This action will ensure that the labwareData state aligns consistently with the barcodes in the input field.

src/lib/receptions/SequencescapeTubes.js
stevieing: Could the default have both plates and tubes and then each reception uses the appropriate one?

BenTopping (commented):
Looking good, couple comments.

src/lib/receptions/SequencescapeTubes.js
seenanair: That sounds good. I will  make the changes  👍

src/views/GeneralReception.vue
seenanair: Yes sure. Do you have some time tomorrow after 10.30 , for a quick call?

src/views/GeneralReception.vue
```diff
@@ -272,6 +272,7 @@ export default {
       attributes: [], // attributes is used to store the attributes that have been found in sequencescape
     },
     printerName: '',
+    debounceTimer: 1000, // debounce timer for barcode deletion
```
BenTopping, line 275: I think this should be set to null by default since its not the time its the timer itself. I think thats my mistake when we went over it.

BenTopping (dismissed):

src/views/GeneralReception.vue
seenanair: @BenTopping 
No, you helped a lot. It's bit stupid of me that I didn't realise that it's a timerId returned by setTimeout and not a timer value.
Updated to null now 👍

BenTopping (approved):

stevieing (approved):

PR #1469, opened 2024-01-03
DPL-987 - Pacbio Library request changes

BenTopping:

- Changes pacbio library requests so that the tube comes from the pool. This is done in preparation for the aliquot work.

Linked service PR: https://github.com/sanger/traction-service/pull/1174

### DIFF
--- a/src/store/traction/pacbio/libraries/actions.js
+++ b/src/store/traction/pacbio/libraries/actions.js
@@ -50,7 +50,7 @@ const setLibraries = async ({ commit, getters }, filter, page) => {
   const promise = request.get({
     page,
     filter,
-    include: 'request,tag,tube,pool',
+    include: 'request,tag,pool.tube',
   })
   const response = await handleResponse(promise)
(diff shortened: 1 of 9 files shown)


src/store/traction/pacbio/libraries/actions.js
```diff
@@ -82,63 +83,6 @@ const setLibraries = async ({ commit, getters }, filter, page) => {
   return { success, errors, meta }
 }
 
-const updateTag = async ({ getters }, payload) => {
```
stevieing, line 85: I presume these methods are no longer used?

stevieing (approved):

src/store/traction/pacbio/libraries/actions.js
BenTopping: Yeah they aren't used. I think they were from when you could edit a library individually and not through a pool, but it would be best to remove them to prevent confusion.

PR #1497, opened 2024-01-19
Pinia store creation for pacbio/libraries

seenanair:
Closes #

- 

### DIFF
--- a/src/stores/pacbioLibraries.js
+++ b/src/stores/pacbioLibraries.js
@@ -1,0 +1,132 @@
+import { defineStore } from 'pinia'
+import useRootStore from '@/stores'
+import { handleResponse } from '@/api/ResponseHelper.js'
+import { groupIncludedByResource, dataToObjectById } from '@/api/JsonApi.js'
+import handlePromise from '@/api/PromiseHelper.js'
+
+export const usePacbioLibrariesStore = defineStore('pacbioLibraries', {
+  state: () => ({
+    tagSetChoices: [],
+    tagChoices: {},
+    libraries: [],
+  }),
+
+  getters: {
+    libraryRequest: () => {
+      const rootStore = useRootStore()
+      return rootStore.api.traction.pacbio.libraries
+    },
+    requestLibraryRequest: () => {
+      const rootStore = useRootStore()
+      return rootStore.api.traction.pacbio.request_library
+    },
+    // Returns a list of tag options by tagSetId
+    tagChoicesForId: (state) => (tagSetId) => {
+      return state.tagChoices[tagSetId] || []
+    },
+  },
+  actions: {
+    async createLibraryInTraction(library) {
+      const rootState = useRootStore()
+      // Some duplication of code from createPool but this is for single library pool
+      const tag_id = library.tag.id
+      const body = {
+        data: {
+          type: 'pools',
+          attributes: {
+            library_attributes: [
+              {
+                pacbio_request_id: library.sample.id,
+                template_prep_kit_box_barcode: library.template_prep_kit_box_barcode,
+                tag_id,
+                volume: library.volume,
+                concentration: library.concentration,
+                insert_size: library.insert_size,
+              },
+            ],
+            template_prep_kit_box_barcode: library.template_prep_kit_box_barcode,
+            volume: library.volume,
+            concentration: library.concentration,
+            insert_size: library.insert_size,
+          },
+        },
+      }
+      const request = rootState.api.traction.pacbio.pools
+      const promise = request.create({ data: body, include: 'tube' })
+      const { success, data: { included = [] } = {}, errors } = await handleResponse(promise)
+      const { tubes: [tube = {}] = [] } = groupIncludedByResource(included)
+      const { attributes: { barcode = '' } = {} } = tube
+      return { success, barcode, errors }
+    },
+    async deleteLibraries(libraryIds) {
+      const promises = this.libraryRequest.destroy(libraryIds)
+
+      const responses = await Promise.all(promises.map((promise) => handlePromise(promise)))
+      return responses
+    },
+    async setLibraries(filter, page) {
+      const request = this.libraryRequest
+      const promise = request.get({
+        page,
+        filter,
+        include: 'request,tag,pool.tube',
+      })
+      const response = await handleResponse(promise)
+
+      const { success, data: { data, included = [], meta = {} } = {}, errors = [] } = response
+      const { tubes, tags, requests, pools } = groupIncludedByResource(included)
+
+      if (success) {
+        /* 
+      Here we build library objects to include necessary relational data
+      for the pacbio libraries page
+    */
+        const libraries = data.map((library) => {
+          const pool = pools?.find((pool) => pool.id == library.relationships.pool.data?.id)
+          return {
+            id: library.id,
+            ...library.attributes,
+            pool,
+            tag_group_id: tags?.find((tag) => tag.id == library.relationships.tag.data?.id)
+              ?.attributes.group_id,
+            sample_name: requests?.find(
+              (request) => request.id == library.relationships.request.data?.id,
+            )?.attributes.sample_name,
+            barcode: tubes?.find((tube) => tube.id == pool.relationships.tube.data?.id)?.attributes
+              .barcode,
+          }
+        })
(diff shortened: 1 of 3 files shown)


src/stores/pacbioLibraries.js
```diff
@@ -0,0 +1,143 @@
...
+  state: () => ({
+    tagSetChoices: [],
+    tagChoices: {},
+    libraries: [],
+  }),
+
+  getters: {
+    /**Return all libraries */
+    libraryRequest: () => {
+      const rootStore = useRootStore()
```
BenTopping, line 20: Can you initialise useRootStore() at the top of the file? Instantiated in quite a few places and its not dynamic.

Does libraryRequest need to be a getter? Seems to only be used once or twice in the actions so not sure if its need to be a getter.

src/stores/pacbioLibraries.js
```diff
@@ -0,0 +1,143 @@
...
+      return rootStore.api.traction.pacbio.libraries
+    },
+    // Return the library requests
+    requestLibraryRequest: () => {
```
BenTopping, line 24: Can't see any references to requestLibrayRequest so can be removed?

src/stores/pacbioLibraries.js
```diff
@@ -0,0 +1,143 @@
...
+            .map(({ id: value, group_id: text }) => ({ value, text }))
+        })
+        this.tagSetChoices = tagSetChoices
+        this.tagChoices = tagChoices
+      }
+      return { success, errors, response }
+    },
+
+    /**Update a library */
+    async updateLibrary(library) {
```
BenTopping, line 136: Is this method still used? Might be able to be removed as well

BenTopping (commented):
Looks like a good start. Are you planning to replace the old vuex store with this one as well? Might also be good to try and use composition API with the library creation component and page?

seenanair:
> Looks like a good start. Are you planning to replace the old vuex store with this one as well? Might also be good to try and use composition API with the library creation component and page?

Replacing of old vuex store is not part of this story. Until the library creation is ready on core side, we will be using old vuex store, I think

src/stores/pacbioLibraries.js
seenanair: God point. I will have a check 👍

src/stores/pacbioLibraries.js
seenanair: For the useRootStore comment, I am afraid it is not possible with this way of  defining stores  which  is called as 'option stores' which defines getters, actions and state.
With option store, we need to call the other store in each action or getter where you need it, though that each call to useRootStore  will return the same instance of the store and not creating a new instance each time you call it.
There is another way of defining stores which are 'setup' stores (which is a very similar concept with options api and  composition apis) which allows it. 
I opted for option stores because I thought that will give a more clear transition from how it is defined in VueX store.

BenTopping: Ah ok makes sense, yes I have only used the setup/options api store so I was in that mind set when looking at this. Thanks for explaining.

src/stores/pacbioLibraries.js
```diff
@@ -0,0 +1,122 @@
...
+      const request = rootState.api.traction.pacbio.libraries
+      const promise = request.create({ data: body, include: 'tube,primary_aliquot' })
+      const { success, data: { included = [] } = {}, errors } = await handleResponse(promise)
+      const { tubes: [tube = {}] = [] } = groupIncludedByResource(included)
+      const { attributes: { barcode = '' } = {} } = tube
+      return { success, barcode, errors }
+    },
+
+    /**Delete all libraries with given library ids */
+    async deleteLibraries(libraryIds) {
```
BenTopping, line 53: Bit of a side note but I think I am against the ability to delete libraries after we make these changes. Since they will be able to be used in runs directly now it makes library data more valuable. Thoughts? @stevieing @seenanair

src/stores/pacbioLibraries.js
```diff
@@ -0,0 +1,122 @@
...
+    /**Fetch all tag sets */
+    async fetchPacbioTagSets() {
+      const rootStore = useRootStore()
+      const promise = rootStore.api.traction.pacbio.tag_sets.get({ include: 'tags' })
+      const response = await handleResponse(promise)
+      const { success, data: { data, included = [] } = {}, errors = [] } = response
+      if (success) {
+        const tagSets = dataToObjectById({ data, includeRelationships: true })
+        const tags = dataToObjectById({ data: included })
+        const tagSetChoices = []
```
BenTopping, line 108: Can we update tagSetChoices and tagChoices directly instead of having a temporary variable now?

tests/unit/stores/pacbioLibraries.spec.js
```diff
@@ -0,0 +1,260 @@
...
+            id: '7',
+            state: 'pending',
+            volume: 1,
+            concentration: 1,
+            template_prep_kit_box_barcode: 'LK12345',
+            insert_size: 100,
+            created_at: '2019/10/16 13:52',
+            deactivated_at: null,
+            source_identifier: null,
+            pool: undefined,
```
BenTopping, line 198: Can remove pool.

BenTopping (commented):
Looks good. Very minor comments.

src/stores/pacbioLibraries.js
```diff
@@ -0,0 +1,122 @@
...
+        filter,
+        include: 'request,tag,tube',
+      })
+      const response = await handleResponse(promise)
+
+      const { success, data: { data, included = [], meta = {} } = {}, errors = [] } = response
+      const { tubes, tags, requests } = groupIncludedByResource(included)
+
+      if (success) {
+        /*
```
stevieing, line 76: I'm not entirely sure what is going on here but it might be worth working through it to use the up to date methods. It looks like we are returning all of the libraries and then looping through all of them to add relationships. What we would do now is put all of the relationships in the store and then merge them together when the user wants to view a library. In a sense we are doing a lot of unnecessary manipulation.

src/stores/pacbioLibraries.js
```diff
@@ -0,0 +1,122 @@
+import { defineStore } from 'pinia'
```
stevieing, line 1: I think it would be good to provide some clearer documentation to the jsdoc standard?

src/stores/pacbioLibraries.js
stevieing: Again I think we might be doing stuff we don't need to. It looks like we are setting up all of the tag sets when in essence most of the time they won't be using it. It would be good to work through it to see exactly what it going on and also look at what we are doing in the pools page.

stevieing (changes requested):
I think most of the changes I have suggested are related to how it was originally setup and not the move over to Pinia. As we are reworking it it might be a good idea to refactor to our newer standards.

src/stores/pacbioLibraries.js
seenanair: For most of the stores, this seems to be a common pattern, which I thought would be the way the re-formatting is done on client side for the data fetched from store.
 It seems like the current process involves creating a new library object and populating its attributes/fields from relationships in the library object returned by the service.  For eg, a new barcode field is defined for library object which is retrieved from tube->barcode relationship.

I don't have much insight into json api, but just a thought which is possible in  graphql. 
To make it simpler, would it better for the service to handle these complexities of flattening these relations , ensuring that the client receives a simplified object in the expected format?
I might require further inputs to proceed on this.

src/stores/pacbioLibraries.js
seenanair: @stevieing It would be great if we could discuss a few of these comments tomorrow.

src/stores/pacbioLibraries.js
stevieing: This is an example of where we are doing it differently using side loading / lazy loading https://github.com/sanger/traction-ui/blob/develop/src/stores/pacbioRunCreate.js#L246. It is far more efficent and is less complex in the code.

stevieing (approved):
I think we can just merge this directly into develop as it does not affect anything?

seenanair:
> I think we can just merge this directly into develop as it does not affect anything?

@stevieing Yes, it is not used any where. Shall I will merge this?

Issue #1511, opened 2024-01-25
DPL-1066 As developers, we need to modify the libraries page so that it will only list the real libraries (libraries that belong to single-plex pool)


seenanair:
**User story**
 As developers, we need to modify the libraries page so that it will only list the real libraries (libraries that belong to single-plex pool).

**Who are the primary contacts for this story**
James / Steve

**Who is the nominated tester for UAT**
PSD

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Pacbio libraries page should only list libraries that are real library entities (currently belong to single plex pools)
- [ ] update Resolve Operations to improve performance https://vitejs.dev/guide/performance.html

**Dependencies**
This story is blocked by the following dependencies:
- sanger/traction-service#1137

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

Issue #1509, opened 2024-01-25
DPL-1064 As developers, we need to modify the traction UI to use the Pinia store for libraries instead of the Vuex store to support revio multiplexing


seenanair:
**User story**
As developers, we need to modify the traction UI to use the Pinia store for libraries instead of the Vuex store to support revio multiplexing

**Who are the primary contacts for this story**
James/Steve

**Who is the nominated tester for UAT**
PSD

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] All library related UI components should be using pinia store for libraries
- [ ] Removal of vuex library store
- [ ] Use composition API
- [ ] update Resolve Operations to improve performance https://vitejs.dev/guide/performance.html

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/traction-ui#1421

References
This story has a non-blocking relationship with:
- sanger/traction-service#1137

**Additional context**

PR #1535, opened 2024-02-13
[EPIC] Library creation for multiplexing

seenanair:
Changes proposed in this pull request

DPL-1064 (https://github.com/sanger/traction-ui/issues/1509)
DPL-1066 (https://github.com/sanger/traction-ui/issues/1511)

- Modify PacbioLibraryCreate, PacbioLibraryIndex, and PacbioSampleIndex components to use pacbioLibraries store (Pinia)which supports new library creation.
- Modify PacbioLibraryIndex and PacbioLibraryCreate to use the Composition API.
- Modify libraries page so that it will only list single-plex libraries (which will automatically happen as part of using new store)
- Remove VueX Pacbio/Libraries store.


src/composables/useModalHelper.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * A composable function that provides a method to hide a modal.
+ * @returns {Function} .hide A method that hides the modal.
+ */
+export default function useModalHelper(modalRef) {
+  /**
+   * Hides the modal.
+   */
+  const hide = () => {
```
BenTopping, line 9: I would change this to hideModal as when its used in a component 'hide' is ambiguous.

src/composables/useModalHelper.js
BenTopping, line 1: Can we now remove the ModalHelper mixin in favour of this?

src/composables/useAlert.js
```diff
@@ -0,0 +1,19 @@
+import store from '@/store'
+/**
+ * A composable function that provides a method to show an alert.
+ * @returns {Function} .showAlert A method that shows an alert. It commits a mutation to the Vuex store.
+ * @param {string} message - The message to show in the alert.
+ * @param {string} type - The type of the alert.
+ * @param {string} dataType - The data type of the alert.
+ */
+export default function useAlert() {
+  //const store = useStore()
```
BenTopping, line 10: remove commented out code.

src/stores/pacbioLibraries.js
```diff
@@ -150,7 +152,6 @@ export const usePacbioLibrariesStore = defineStore('pacbioLibraries', {
     async deleteLibraries(libraryIds) {
```
BenTopping, line 152: Libraries should not be destroyable in the UI so we should remove this and the associated components.

src/views/pacbio/PacbioLibraryIndex.vue
```diff
@@ -57,125 +57,116 @@
...
-import DataFetcher from '@/components/DataFetcher'
-import { mapActions, mapGetters } from 'vuex'
-import { getCurrentDate } from '@/lib/DateHelpers'
-import useQueryParams from '@/lib/QueryParamsHelper'
+<script setup>
+import PrinterModal from '@/components/PrinterModal.vue'
+import FilterCard from '@/components/FilterCard.vue'
+import DataFetcher from '@/components/DataFetcher.vue'
+import { getCurrentDate } from '@/lib/DateHelpers.js'
+import useQueryParams from '@/lib/QueryParamsHelper.js'
```
BenTopping, line 65: Is it worth moving useQueryParams form src/lib to src/composables not that composables has a designated folder?

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -107,80 +107,97 @@
...
+const modalRef = ref(null)
+
+//Composables
+const { showAlert } = useAlert()
+const { hide } = useModalHelper(modalRef.value)
+
+// Define emits
+const emit = defineEmits(['alert'])
+//Emits the 'alert' event
+const emitAlert = (message) => {
```
BenTopping, line 141: This emit also needs to emit the alert type. Currently if you create a library you get a default gray alert colour but it should be green ('success'). I believe this is where it is getting blocked from.

BenTopping: You can probably remove this emitAlert method and call emit('alert') directly where its used as it is only used once.

BenTopping (commented):
Couple minor comments. Overall looks really good 👌 .

src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -107,80 +107,97 @@
...
+import { computed, ref } from 'vue'
+import { usePacbioLibrariesStore } from '@/stores/pacbioLibraries'
+import useAlert from '@/composables/useAlert.js'
+import useModalHelper from '@/composables/useModalHelper.js'
+
+// Define props
+const props = defineProps({
+  disabled: Boolean,
+  isStatic: Boolean,
+  selectedSample: {
```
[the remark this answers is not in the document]
seenanair: Yes, I will change it 👍

src/composables/useModalHelper.js
seenanair: ModalHelper is used in one other place (PacbioSampleMetaEdit). So it is better to remove whenever we refactor that component, I hope.

src/stores/pacbioLibraries.js
stevieing: I think libraries should be destroyable as long as they have not been used in a pool or a sequencing run. Also I am not sure whether we have an active status on libraries which could be another option.

stevieing (commented):
My only comment would be that it is good to ensure that there is a good level of documentation wherever possible especially where we are doing new stuff that will be reused elsewhere.

src/stores/pacbioLibraries.js
BenTopping: OK. Worth discussing. If they can be destroyable under certain conditions we will need a story for it. We will need to implement before this epic branch goes to prod because otherwise users can destroy libraries tied to runs which would break a lot of things.

seenanair:
> My only comment would be that it is good to ensure that there is a good level of documentation wherever possible especially where we are doing new stuff that will be reused elsewhere.

@stevieing Thank you.
 I have tried to put in documentation on composables which I thought is useful as it is going to be reused. If you would like to mention any other files, please mention and I will try to update

src/composables/useModalHelper.js
seenanair: Done

src/stores/pacbioLibraries.js
stevieing: Could you add it to your epic branch. It is a minor change and we just need to check if either it belongs to a sequencing run or a pool? You can use the before_destroy hook.

BenTopping (commented):
I won't approve it yet because this branch is still waiting on other work but your changes look good 👍

stevieing (approved):

PR #1564, opened 2024-03-01
Dpl 996 pacbio pool create, pool stores to pinia

seenanair:
- pacbioPoolCreate, pacbioPools migration to Pinia
- tags and tagsets moved to root store


src/stores/pacbioPoolCreate.js
```diff
@@ -0,0 +1,954 @@
...
+     * // Create a pool
+     * const result = await createPool();
+     * console.log(result); // { success: true, barcode: 'barcode123', errors: [] }
+     */
+    async createPool() {
+      const { libraries, pool } = this
+      if (!validate(libraries)) return { success: false, errors: 'The pool is invalid' }
+      const rootStore = useRootStore()
+      const request = rootStore.api.traction.pacbio.pools
+      const promise = request.create({
```
BenTopping, line 539: Is there a plan here to feature flag between sending used_aliquots vs libraries?

src/stores/pacbioPoolCreate.js
```diff
@@ -0,0 +1,954 @@
...
+     * // Create a pool
+     * const result = await createPool();
+     * console.log(result); // { success: true, barcode: 'barcode123', errors: [] }
+     */
+    async updatePool() {
+      const { libraries, pool } = this
+      if (!validate(libraries)) return { success: false, errors: 'The pool is invalid' }
+      const rootStore = useRootStore()
+      const request = rootStore.api.traction.pacbio.pools
+      const promise = request.update(payload({ libraries, pool }))
```
BenTopping, line 570: Do we need to feature flag sending used_aliquots vs libraries?

src/stores/pacbioPoolCreate.js
```diff
@@ -0,0 +1,954 @@
...
+      const promise = request.find({
+        id: poolId,
+        // We want to load *all* associated records, as otherwise we might be referencing them
+        // before they are loaded. Furthermore, if we start filtering the plates list at all,
+        // we may *never* load the relevant records.
+        // We load the other wells associated with the plate too, to ensure the remaining plate
+        // doesn't appear empty. This is especially important if the pool request finishes
+        // after the request for all plates, as otherwise the partial record will over-write
+        // the full one.
+        include:
```
BenTopping, line 602: I think the feature flag should flip between pulling back libraries or used_aliquots. If we don't feature flag and used_aliquots doesn't work we would have to rollback in this setup.

src/stores/pacbioPoolCreate.js
```diff
@@ -0,0 +1,954 @@
...
+          'used_aliquots.tag.tag_set,used_aliquots.source_plate.wells.requests,used_aliquots.request.tube,tube',
+      })
+      const response = await handleResponse(promise)
+
+      const { success, data: { data, included = [] } = {}, errors = [] } = response
+
+      if (success) {
+        const {
+          library_pools,
```
BenTopping, line 611: If used_aliquots are pulled back they will need to be added here as they will be an included group

src/stores/pacbioPoolCreate.js
```diff
@@ -0,0 +1,954 @@
...
+        } = groupIncludedByResource(included)
+        // Get the pool tube and remove it from tubes list
+        const poolTube = tubes.splice(
+          tubes.indexOf((tube) => tube.id == data.relationships.tube.data.id),
+          1,
+        )[0]
+        //Populate pool attributes
+        this.pool = { ...data }
+        //Populate libraries
+        const newLibraries = dataToObjectById({ data: library_pools, includeRelationships: true })
```
BenTopping, line 626: If we pull back libraries we will have library_pools but if we pull back used_aliquots the data will need to be used_aliquots

BenTopping (changes requested):
I think now the service is ready it would be useful to test how this behaves with used_aliquots vs libraries. One caveat I can see is used_aliquots don't have a pacbio_request_id they have a source_id. We could edit used_aliquots when we retrieve them to set pacbio_request_id as source_id.

I think we also need to consider if/how we feature flag used_aliquots vs libraries. If we feature flag we will need to change how we retrieve pools, update and create pools.

seenanair:
> This is a good start.
> 
> I think now the service is ready it would be useful to test how this behaves with used_aliquots vs libraries. One caveat I can see is used_aliquots don't have a pacbio_request_id they have a source_id. We could edit used_aliquots when we retrieve them to set pacbio_request_id as source_id.
> 
> I think we also need to consider if/how we feature flag used_aliquots vs libraries. If we feature flag we will need to change how we retrieve pools, update and create pools.

Thank you. I think setting source_id can be done as part of store action.

src/stores/index.js
```diff
@@ -3,16 +3,97 @@ import build from '@/api/ApiBuilder'
...
+      const request = this.api.traction.tags
+      const promise = request.get()
+      const {
+        success,
+        data: { data, included = [] },
+        errors,
+      } = await handleResponse(promise)
+
+      if (success && data) {
+        const { tags } = groupIncludedByResource(included)
```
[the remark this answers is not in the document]
seenanair: Done

src/stores/pacbioLibraries.js
```diff
@@ -107,25 +54,26 @@ export const usePacbioLibrariesStore = defineStore('pacbioLibraries', {
...
           which may not happen in all the places where it's called. 
           Hence, a search in both places is required to ensure that librariesArray returns the correct tag 
           associated with all libraries."*/
-
           const tagGroupId = tagId
-            ? state.libraryState.libraryTags[tagId]
-              ? state.libraryState.libraryTags[tagId].group_id
-              : state.tagState.tags[tagId]
-                ? state.tagState.tags[tagId].group_id
+            ? state.libraryTags[tagId]
```
stevieing, line 73: Do you need to use a nested ternary here? Can you do the question mark after [tagId] ...

src/stores/pacbioPools.js
```diff
@@ -0,0 +1,110 @@
...
+     * returns the pools with the retrieved data.
+     * The used_aliquots for each pool are retrieved by mapping over the `pool.used_aliquots`, retrieving the used_aliquot data from `state.used_aliquots`
+     * based on the used_aliquot ID, and returning the used_aliquot data with the sample name from `state.requests` and the group ID from `state.tags`.
+     * The barcode for each pool is retrieved from `state.tubes` based on the `pool.tube`.
+     * The run suitability for each pool is retrieved from `pool.run_suitability` and formatted with the `buildRunSuitabilityErrors` function.
+     *
+     * @param {Object} state - The state object.
+     * @returns {Object[]} The array of pools with the retrieved data.
+     */
+    poolsArray: (state) => {
```
stevieing, line 46: It looks like we are doing objectArray in a lot of places. There might be an opportunity for abstraction somewhere.

src/stores/utilities/pool.js
```diff
@@ -0,0 +1,114 @@
...
+ * @returns {boolean} Returns true if all used_aliquots are valid and there are no duplicate tags, false otherwise.
+ *
+ * @example
+ * const used_aliquots = {
+ *   '1': { tag_id: 'tag1', volume: 10, concentration: 5, insert_size: 1000,source_id:'1' },
+ *   '2': { tag_id: 'tag2', volume: 10, concentration: 5, insert_size: 1000,source_id:'1'},
+ * };
+ * const isValid = validate(used_aliquots); // returns true
+ */
+const validate = (used_aliquots) => {
```
stevieing, line 58: Again wondering if there is any opportunity for refactoring. It looks like we are looping through the entries 3 times.

src/stores/utilities/pool.js
```diff
@@ -0,0 +1,114 @@
...
+ * { data: { type: 'pools', attributes: { used_aliquot_attributes: [ used_aliquot1, used_aliquot2 ... ], template_prep_kit_box_barcode, volume, concentration, insert_size}}}
+ */
+const payload = ({ used_aliquots, pool }) => {
+  const { template_prep_kit_box_barcode, volume, concentration, insert_size } = pool
+  return {
+    data: {
+      type: 'pools',
+      id: pool.id,
+      attributes: {
+        used_aliquots_attributes: Object.values(used_aliquots).map((used_aliquot) =>
```
stevieing, line 94: Not entirely sure what is going on here. There is an underscore then .pick?

stevieing (changes requested):
Some things to ponder on.

src/stores/pacbioLibraries.js
seenanair: Good point, the outer one can very well be removed as we are checking state.libraryTags[tagId].

src/stores/utilities/pool.js
seenanair: I have refactored it, so that as soon something is found wrong it will set value to false, so it avoids one round of looping.

src/stores/utilities/pool.js
seenanair: pick is a loadash function ( _ seems to be. a common convention/practice to denote a lodash library instance) which is  used to return a copy of the object that is composed of the picked object properties. This reduces the wordiness/verbosity of spreading the object for using required attributes

src/stores/utilities/pool.js
seenanair: Another possibility is to avoid looping through the entire object if something is wrong. But ,I think this will be helpful for user to know all the errors associated with all fields in one go.

src/stores/pacbioPools.js
stevieing: I understand but it is something worth thinking about. Javascript is a functional language so is built for stuff like this. Sounds like we are formatting all fetched data to a shape that components can access so we can pass functions to do certain parts of it.

src/stores/pacbioRoot.js
```diff
@@ -0,0 +1,91 @@
...
+     * @function fetchPacbioTagSets
+     * @returns {Promise<{success: boolean, errors: Array, response: Object}>} - A promise that resolves to an object containing a success boolean, or an error string.
+     * @throws {Error} - Throws an error if the request fails.
+     */
+    async fetchPacbioTagSets() {
+      const rootStore = useRootStore()
+      const promise = rootStore.api.traction.pacbio.tag_sets.get({ include: 'tags' })
+      const response = await handleResponse(promise)
+      const { success, data: { data, included = [] } = {}, errors = [] } = response
+      if (success && data.length > 0) {
```
stevieing, line 84: Something also worth considering is that for a fetch we always check that data.length is greater than 0 so it would be worth changing success to automatically include it for get requests.

src/stores/utilities/pool.js
stevieing: That makes sense.

src/stores/utilities/pool.js
stevieing: OK. That's fine but worth commenting as such. I wouldn't be too concerned about lodash as it is a well supported library but it is always worth keeping an eye on the external libraries we are using that are just providing syntactic sugar e.g. if we start using pick quite regularly it is worth centralising it's usage and centralising (as with Axios) just in case it becomes unsupported.

src/stores/pacbioRoot.js
seenanair: Yes,  good point. Probably we can add this as part of dpl-1093 story

seenanair:
> Looks good 👌 May be a few things we need to tweak when we come to implementation but this seems as good as it can be considering the unknowns.

Thank you very much for your feedback. 
Yes, sure. We can always come back to this and modify (if required) as part of component stories I hope.

PR #1572, opened 2024-03-06
Y24-02  Pacbio pools page using pacbio_pools Pinia store

seenanair:
Following are the changes proposed in this pull request

- Created Pacbio Pools page that uses pacbio_pools Pinia store which supports used_aliquots
- Feature flag support (multiplexing_phase_2_aliquot) to toggle between the Pacbio pools page that uses libraries (VueX store) and the one that uses used_aliquots (Pinia).


src/router.js
```diff
@@ -130,7 +130,7 @@ const router = createRouter({
         {
           path: 'pools',
           name: 'PacbioPoolIndex',
-          component: PacbioPoolIndex,
+          component: PacbioPoolIndexToggleWithFeatureFlag,
```
BenTopping, line 133: Instead of a separate component to handle the feature flag could we handle it in here? So the router decides which component to show?
Im not sure if it is possible because its asynchronous but might be cleaner.

src/views/pacbio/PacbioPoolIndexToggleWithFeatureFlag.vue
BenTopping, line 1: If we go with the approach of a separate component for feature flagging can we make it generic? Is it possible to make a view called 'FeatureFlagView' and we pass it the flag we are checking and the two views as props?

src/views/pacbio/PacbioPoolIndex.vue
```diff
@@ -60,7 +60,7 @@
 
         <template #row-details="row">
           <div>
-            <traction-table :items="row.item.libraries" :fields="field_in_details">
+            <traction-table :items="row.item.used_aliquots" :fields="field_in_details">
```
BenTopping, line 63: Tested this and I don't think this is working properly.
This shows we are missing some tests for this behaviour as well. Maybe worth adding some to the cypress tests for this page.

__mocks__/swrv.js
```diff
@@ -2,6 +2,9 @@ const baseURL = import.meta.env.VITE_TRACTION_BASE_URL
 
 const flags = {
   flipper_id: 'User',
+  features: {
```
BenTopping, line 5: Where are these used?
Maybe worth adding some documentation here because I'm not sure what is accessing the feature flags through this.

BenTopping (changes requested):
Small issue with 'Show details' on the pools page.
The rest looks good. Just some comments/questions about the feature flag design.

src/views/pacbio/PacbioPoolIndexToggleWithFeatureFlag.vue
BenTopping: Maybe an extension of the featureflagged component

src/router.js
seenanair: I had a look into it and we can probably use router navigation guards like 'beforeEach' , but that will called for every routing which I thought won't be appropriate.

src/views/pacbio/PacbioPoolIndex.vue
seenanair: Good finding. I will have a look.

src/router.js
seenanair: To make this work, I think we should have a way to query feature flag status. Currently we are managing it through useSWRVv from swrv library., but that seems to work only within a component context it seems.

src/views/pacbio/PacbioPoolIndex.vue
seenanair: Fixed

__mocks__/swrv.js
seenanair: You are right. It is not required, it seems which I thought is used  while mocking

src/router.js
stevieing: Sounds like we need to look into this. I was never sure about the use of swrv as it is another dependency.

BenTopping (approved):
Looks good. Did you find any docs about passing in Vue components as props? Struggled to find any online. Just want to make sure its not an anti pattern.

PR #1582, opened 2024-03-14
Dpl 993 pacbio pool create

stevieing:
This PR includes -
Feature flag support.
Duplication of all Pacbio Pool edit and pool create components to separate VueX and Pinia stores using feature flags.
Refactoring of all Pacbio Pool edit and pool create components to use Pinia store.
Refactoring of all components to use the composition API.
UI improvements for Pacbio Pool create/edit page.
Auto-filling of aliquot fields for all library requests in pool creation page.
Store changes to support aliquots and data cleanup.
Updates to unit and e2e test cases and changes to test data to support aliquots.


src/stores/index.js
```diff
@@ -87,6 +88,8 @@ const useRootStore = defineStore('root', {
      */
     addCSVLogMessage(info, error, type = 'danger') {
       this.addMessage({ type, message: errorFor(info, error) }, { root: true })
```
BenTopping, line 90: Is line 90 still needed?

src/stores/pacbioPoolCreate.js
```diff
@@ -268,14 +280,15 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
-
     /**
      * Retrieves the pool item from the state. If the pool item does not exist, returns an empty object.
      *
      * @param {Object} state - The state object that contains the pool item.
      * @returns {Object} The pool item from the state, or an empty object if it does not exist.
      */
-    poolItem: (state) => state.pool || {},
+    poolItem: (state) => {
+      return state.pool || {}
```
BenTopping, line 290: Does state.pool not default to {} anyway? Could we update the places we are referencing poolItem to just get the pool from state and remove this function?

src/stores/pacbioPoolCreate.js
```diff
@@ -492,15 +508,15 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
      *
      * @example
      * // Deselect a tube and its contents
-     * deselectTubeAndContents('TRAC-123')
+     * deselectTubeAndContents('1')
      */
-    deselectTubeAndContents(tubeBarcode) {
-      const tube = Object.values(this.resources.tubes).find((tube) => tube.barcode == tubeBarcode)
+    deselectTubeAndContents(id) {
+      const tube = Object.values(this.resources.tubes).find((tube) => tube.id == id)
```
BenTopping, line 519: If tube resources are indexed by id we can just do `this.resources.tubes[id]`

src/stores/pacbioPoolCreate.js
```diff
@@ -786,7 +860,7 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
 
       // We will be return a successful empty list if no plates match the filter
       // Therefore we want to return an error if we don't have any plates
-      if (!data.length) {
+      if (!data || !data.length) {
```
BenTopping, line 863: `!data?.length`

src/stores/pacbioPoolCreate.js
```diff
@@ -916,7 +1032,24 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
 
     selectRequest({ id, selected = true }) {
       if (selected) {
-        this.used_aliquots[`_${id}`] = createUsedAliquot({ source_id: id })
+        //If the libraries contain the request, fill the used_aliquot values with the library attributes values for template_prep_kit_box_barcode, volume, concentration, and insert_size
+        const library = Object.values(this.resources.libraries).find(
```
BenTopping, line 1036: If libraries are indexed by ID we can do `this.resources.libraries[library.request]`

src/stores/pacbioPoolCreate.js
```diff
@@ -916,7 +1032,24 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
+          (library) => library.request == id,
+        )
+        const request = this.resources.requests[id]
+        const libraryAttributes = library
```
BenTopping, line 1040: Not entirely sure whats going on here

src/views/pacbio/PacbioPoolCreate.vue
```diff
@@ -1,39 +1,36 @@
...
-        </div>
-        <div v-else>
-          <traction-section title="Tubes" number="2a">
-            <div class="text-left">Find Tubes</div>
-            <LabwareFinder :fetcher="findPacbioTube" filter="barcode" />
+        <div class="flex flex-col">
+          <traction-section
+            title="Scan labware"
+            number="1a"
+            description="To get started, scan or type a plate, tube, or library barcode and press Enter"
```
BenTopping, line 9: Should be keep library barcode in description or wait until we fully add it? Don't want to confuse users.

BenTopping (changes requested):
Works well and looks good 👌  Couple syntax comments and questions. Quite a big PR so I will give it another look over later but hopefully I covered most of the issues/questions.

seenanair:
> Works well and looks good 👌 Couple syntax comments and questions. Quite a big PR so I will give it another look over later but hopefully I covered most of the issues/questions.

Thank you. Please do more checks if you feel like as it is a big one. 👍

src/components/labware/PacbioTubeWell.vue
```diff
@@ -0,0 +1,67 @@
...
+   */
+  requests: {
+    type: Object,
+    required: true,
+  },
+  /*
+   * The position of the well
+   * If the position is provided, it will be displayed in the well
+   */
+  position: {
```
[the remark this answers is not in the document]
seenanair: It is not currently in use, but it is kept there in case it is required for any future context with a position text displayed. However, yes, it can be removed. 👍

src/stores/index.js
seenanair: Yes, eventually we need to remove the Vuex store for messages (traction/addMessage - line 92) and use this store, so it's probably good to keep the Pinia store functioning.

src/stores/pacbioPoolCreate.js
seenanair: Yes, upon checking, it seems the library's ID is different from library.request. So, we might not be able to access it this way, which could result in retrieving the wrong library object.

seenanair: Yes, but on check it seems libraries id is different from library.request. So, we might not be able to access it this way which might return a wrong library object

src/stores/pacbioPoolCreate.js
seenanair: While selecting a request and the request is associated with a library, we are filling the used_aliquot values with the library's attribute values for template_prep_kit_box_barcode, volume, concentration, and insert_size. This is useful for auto-populating selected request information on the pool creation page.
There is a comment on top which describes it. Hope it make sense

src/views/pacbio/PacbioPoolCreate.vue
seenanair: Good finding! We don't need it now, as it will confuse user 👍

src/router.js
```diff
@@ -162,8 +163,12 @@ const router = createRouter({
         {
           path: 'pool/:id',
           name: 'PacbioPoolCreate',
-          component: PacbioPoolCreate,
-          props: true,
+          component: FlaggedFeatureView,
+          props: {
+            feature: 'multiplexing_phase_2_aliquot',
```
BenTopping, line 168: I would use a different feature flag to this because this one is already in prod and we dont want to turn off existing working features if this one needs to be toggled.

src/stores/pacbioPoolCreate.js
```diff
@@ -251,15 +261,17 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
     requestList: (state) => (ids) => {
       const requests = state.resources.requests
       const selectedRequests = state.used_aliquots
+      let val = []
       if (ids) {
-        return ids.map((id) => {
+        val = ids.map((id) => {
           return { ...requests[id], selected: !!selectedRequests[`_${id}`] }
```
BenTopping, line 267: did you figure out why underscores are used for ids? I assume its because they are stored in the same place as requests in the store and need some way of differentiating them.

src/stores/pacbioPoolCreate.js
BenTopping: As yes its not the library ID we are using.

src/stores/pacbioPoolCreate.js
BenTopping: Ah yes, missed that comment. Might be clearer to manually assign the required attributes instead of filtering the object, not sure.

BenTopping (changes requested):

src/icons/BarcodeIcon.vue
BenTopping, line 1: Did you not like the old one 😆

seenanair: Felt like it was bit blurry 😁

src/stores/pacbioPoolCreate.js
seenanair: Yes, I believe it is used to differentiate between the ones available and ones selected, I believe,  as all of them are stored in same place. The selected ones are having the _ prefix.

src/stores/pacbioPoolCreate.js
```diff
@@ -1025,16 +1024,20 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
       if (selected) {
         /*If the request is associated with a library, fill the used_aliquot values with the library attributes values 
         for template_prep_kit_box_barcode, volume, concentration, and insert_size*/
+        const autoFillLibraryAttributes = [
```
BenTopping, line 1027: Looks a lot clearer, thanks

src/router.js
```diff
@@ -135,7 +135,7 @@ const router = createRouter({
           name: 'PacbioPoolIndex',
           component: FlaggedFeatureView,
           props: {
-            feature: 'multiplexing_phase_2_aliquot',
+            feature: 'multiplexing_phase_2_pool_with_aliquots',
```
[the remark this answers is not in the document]
seenanair: 🤦‍♀️ Corrected

BenTopping (approved):
Looks good to me 💯

PR #1602, opened 2024-03-28
Bug - Fixes table horizontal overflow issues

BenTopping:

- Fixes horizontal overflow issues

It would be good to pull down and test locally as we use the table in a lot of places.
This change moves the table overflow to be contained within the table component, see below.

Before
Screenshot 2024-03-28 at 14 42 01

After
Screenshot 2024-03-28 at 14 45 13

### DIFF
--- a/src/components/shared/table/TractionTable.vue
+++ b/src/components/shared/table/TractionTable.vue
@@ -40,7 +40,7 @@
 -->
 
 <template>
-  <div class="flex">
+  <div class="flex overflow-x-auto">
     <div class="flex w-full py-2 align-middle inline-block min-w-full">
       <div class="flex flex-col w-full sm:rounded-lg">
         <table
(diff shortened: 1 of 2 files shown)


seenanair (approved):
Working as expected 👍 
But, I have concern about usability, particularly in situations where there are more table rows than can fit within the screen height. When users need to check columns that require horizontal scrolling, it may be challenging for them to navigate to the bottom of the table and then scroll horizontally. This can result in users needing to use both scrollbars simultaneously. 
But again, this can be only evaluated by the users only. So, we can give it a go and see if we are getting any feedback on it. 👍

BenTopping:
Marking as draft for the time being (will re-assess when I get a chance)

Per @seenanair's comments: The current issue is the scroll bar is now pushed to the bottom of the table instead of being on the bottom of the window. This makes the table slighly less useable as to move across a table you first need to scroll to the bottom. There are a few options:

1. Add an artificial scroll bar which sticks to the bottom of the window and scrolls. 
2. Add an additional scrollbar to the top of the table. (doesn't improve the middle of the table much)
3. Do nothing - In most use views the tables don't require horizontal scrolling so the impact should be minimal.

BenTopping:
Closing as this will require more significant work and testing than initially thought since the table is used in a range of places. This would require its own story but its not a priority at this time.

PR #1623, opened 2024-04-15
Y24-050 - Enable libraries to be used in PacBio pools

### LINKED ISSUE #1611: Y24-050 As developer we need to enable the feature flag to add libraries to pool as part of Revio multiplexing
**User story**
As developer we need to enable the feature flag to add libraries to pool as part of Revio multiplexing

**Acceptance criteria**
To be considered successful the solution must include:
- [ ] Enable feature flag multiplexing_phase_2_add_libraries_to_pool
- [ ] Add e2e test case for populating pool request fields if a request from library is selected
- [ ] Add 'pools.used_aliquots.library.request' in method findPoolsOrLibrariesByTube in pacbioRunCreate.js

BenTopping:
Closes #1611 

- Adds test cases for libraries in pools
- Further updates pacbioPoolCreate store to support libraries in pools


src/stores/pacbioPoolCreate.js
```diff
@@ -668,6 +648,27 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
+
+        // Set the used_aliquots
+        Object.values(usedAliquots).forEach((usedAliquot) => {
+          // Creates a request association for the used aliquot based on the source type
+          usedAliquot.request =
+            usedAliquot.source_type == 'Pacbio::Request'
+              ? this.resources.requests[usedAliquot.source_id].id
+              : this.resources.libraries[usedAliquot.source_id].request
+
+          const key = `_${usedAliquot.request}`
```
BenTopping, line 665: Im not entirely sure on the best way to store these selected aliquots. Currently we are using the aliquot's request but this comes with the issue that if the same request is wanting to be used twice, e.g. in a library and a plate then it will cause issues. We have the same issue if we use source_id instead as multiple plates/tubes/libraries may have the same id.

BenTopping: Maybe not using a special id key at all and tracing aliquots back to their well/tube via a mix of the aliquot source_type and its request.

src/stores/pacbioPools.js
```diff
@@ -46,8 +47,13 @@ export const usePacbioPoolsStore = defineStore('pacbioPools', {
     poolsArray: (state) => {
       return Object.values(state.pools).map((pool) => {
         const used_aliquots = pool.used_aliquots.map((used_aliquotId) => {
-          const { id, type, source_id, tag, run_suitability } = state.used_aliquots[used_aliquotId]
-          const { sample_name } = state.requests[source_id] || {}
+          const { id, type, source_id, source_type, tag, run_suitability } =
+            state.used_aliquots[used_aliquotId]
+          // Get the sample name based on the source_type
```
BenTopping, line 52: I have assumed for libraries they want the libraries sample_name rather than just the library barcode.

stevieing (dismissed):
All seems good to me. Just a couple of things (which might come out in UAT):
- do we need to add a flag to the tube to identify that it is a library?
- do we need to add the library barcode to the pooled samples so they can distinguish or is the sample name enough?
- do we need to modif the heading to be pooled aliqouts rather than pooled samples?

seenanair (commented):
Really good work. You have done a lot of cleanup. Well done 👏

seenanair (approved):

PR #1679, opened 2024-05-02
Dpl-1076 - Automatic volume check when library is added to pool 

seenanair:
This PR includes:

Displaying a "available volume" label when the user add a library to pool.
Displaying an error message if the auto-filled (original) volume is greater than available volume
Display error messages when user edits an invalid value to template_kit_box_barcode, volume, insert_size, concentration fields


src/components/pacbio/PacbioPoolAliquotEdit.vue
```diff
@@ -217,4 +228,10 @@ const isValidationExists = (attribute) => {
     return !fieldsThatRequireValidation.value[attribute]
   }
 }
+/*
+  This is to display any volume error for a library request on mounting component in case 
+  the auto-filled volume (original volume of library) is greater than the available volume
+ */
+aliquot.value.source_type === 'Pacbio::Library' &&
```
BenTopping, line 235: Could this be done as part of selectRequest in the store instead? i.e as you select a request we validate it to check the volume and add the error.

src/stores/pacbioPoolCreate.js
```diff
@@ -1080,5 +1088,38 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
+     * @param {Object} used_aliquot_obj - The used aliquot object to validate.
+     * @param {string} field - The field to validate.
+     * @param {any} value - The value to validate is optional.If not provided, the value from the used_aliquot object is used.
+     * @returns {void}
+     *
+     * @example
+     * validateUsedAliquot({ request: 'request1' }, 'field1', 'value1');
+     */
+    validateUsedAliquot(used_aliquot_obj, field, value) {
+      // If the given used_aliquot object is not valid or does not have a 'source_id'property, return without doing anything.
```
BenTopping, line 1104: "does not have a valid 'request' property"

src/stores/utilities/pool.js
```diff
@@ -94,6 +95,37 @@ const validate = ({ used_aliquots, pool }) => {
...
+  const valueToCheck = value != null ? value : used_aliquot[field]
+  if (!valueToCheck) {
+    return 'must be present'
+  }
+  if (
+    field === 'volume' &&
+    used_aliquot.available_volume &&
+    valueToCheck > used_aliquot.available_volume
+  ) {
+    return 'must be less than available volume'
```
BenTopping, line 125: less or equal to

src/stores/utilities/pool.js
```diff
@@ -94,6 +95,37 @@ const validate = ({ used_aliquots, pool }) => {
...
+ * console.log(validateFieldForUsedAliquot(aliquot, 'volume', 10)); // Outputs: ''
+ */
+const validateFieldForUsedAliquot = (used_aliquot, field, value) => {
+  const valueToCheck = value != null ? value : used_aliquot[field]
+  if (!valueToCheck) {
+    return 'must be present'
+  }
+  if (
+    field === 'volume' &&
+    used_aliquot.available_volume &&
```
BenTopping, line 122: if available_volume is 0 this will return false which I dont think is correct.

src/components/pacbio/PacbioPoolAliquotEdit.vue
```diff
@@ -40,8 +40,19 @@
         :with-icon="isValidationExists('volume')"
       >
         <traction-input v-model="volume" data-attribute="volume" placeholder="Volume" />
+        <div class="flex items-center px-1">
+          <traction-tooltip
+            v-show="aliquot.available_volume"
```
BenTopping, line 45: Think this should be something like `aliquot.available_volume != null` otherwise it wont show when available_volume is 0

BenTopping (changes requested):
Works well, one small edge case needs fixing for when available_volume is 0.

BenTopping:
There is also an issue when editing pools. The available volume for an aliquot with an id (exists in the pool already) should be available_volume + aliquot volume as the available volume doesnt take into account the volume you are already using.

seenanair:
>

> There is also an issue when editing pools, the available volume isnt showing up. The existing way of validating volume will also fail when editing an existing aliquot because the available volume doesn't take into account the volume currently used by the aliquot. So you could make available_volume = available_volume + aliquot volume for existing aliquots (aliquots with ids) but I dont know if that will confuse the user to thinking there is more volume than there is. Hopefully that makes sense.

Very Good spot! 👍

src/stores/pacbioPoolCreate.js
```diff
@@ -652,9 +657,18 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
           data: aliquots,
           includeRelationships: true,
         })
-
         // Set the used_aliquots
         Object.values(usedAliquots).forEach((usedAliquot) => {
+          // Creates a request association for the used aliquot based on the source type
```
stevieing, line 662: I think we need a type?

src/stores/pacbioPoolCreate.js
```diff
@@ -1080,5 +1099,38 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
+     *
+     * @param {Object} used_aliquot_obj - The used aliquot object to validate.
+     * @param {string} field - The field to validate.
+     * @param {any} value - The value to validate is optional.If not provided, the value from the used_aliquot object is used.
+     * @returns {void}
+     *
+     * @example
+     * validateUsedAliquot({ request: 'request1' }, 'field1', 'value1');
+     */
+    validateUsedAliquot(used_aliquot_obj, field, value) {
```
stevieing, line 1114: Again this signifies a type might be useful ...

tests/unit/components/pacbio/PacbioLibraryForm.spec.js
```diff
@@ -173,16 +174,10 @@ describe('PacbioLibraryForm.vue', () => {
       expect(wrapper.find('#tag-set-input').element.value).toBe('3')
```
stevieing, line 174: should we use data-attribute?

stevieing (changes requested):
I did run a test locally. It worked but wondering if there should be a sign somewhere of what volume is available? I am not sure if we could do that in real time as we would have to retrieve all of the aliqouts for tht library unless we can add an endpoint in the service to speed it up?

src/stores/pacbioPoolCreate.js
seenanair: Did you mean a type for a volume? We have set different attributes in library resource in service side  and in ui it is set  as fields `available_volume` and `used_volume` in usedAliquot object.

BenTopping:
@stevieing 
There is a service branch you need to see available volume: https://github.com/sanger/traction-service/pull/1306

src/stores/pacbioPoolCreate.js
stevieing: For aliquot. There is a fair amount of logic for validation and checking if it is for a request. We have done this with other objects. It just means that the behaviour is encapsulated and you would be able to reuse it in libraries, pools and sequencing runs.

stevieing (commented):
A couple of things I have noticed:
- When we are editing a pool there is no discernable link between the library and the aliquot so wondering how they know which one they are editing? I guess they can go by the sample name but wondering if it would make sense to have the library barcode in there.
- When you create a pool with a bad volume it just says pool invalid. I can see the error anyway so not too much of a problem but is there a reason we are not showing errors?

src/components/pacbio/PacbioPoolAliquotEdit.vue
```diff
@@ -39,10 +43,15 @@
         :error="errorsFor('volume')"
         :with-icon="isValidationExists('volume')"
       >
-        <traction-input v-model="volume" data-attribute="volume" placeholder="Volume" />
-        <div class="flex items-center px-1">
+        <traction-input
+          v-model="volume"
+          data-attribute="volume"
+          placeholder="Volume"
+          classes="min-w-128"
```
BenTopping, line 50: Is this min-width needed?

src/stores/utilities/usedAliquot.js
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * @param {Object} attributes - The attributes to set on the used aliquot.
+ * @returns {Object} The created used aliquot object.
+ */
+function createUsedAliquot(attributes) {
+  /**
+   * The default attributes for a used aliquot.
+   */
+  const defaultUsedAliquotAttributes = {
+    source_id: null,
```
BenTopping, line 12: source_type?

src/stores/utilities/usedAliquot.js
```diff
@@ -0,0 +1,215 @@
...
+     * @param {boolean} isPool - Whether the aliquot is a pool.
+     * @returns {string[]} An array of required aliquot attributes.
+     *
+     * @example
+     * const attributes = requiredAliquotAttributes(true);
+     * // attributes is ['source_id', 'volume', 'concentration', 'insert_size', 'template_prep_kit_box_barcode', 'tag_id']
+     */
+    requiredAliquotAttributes: function (isPool) {
+      return [
+        'source_id',
```
BenTopping, line 70: source_type required?

BenTopping (commented):
Really nice abstraction!

src/components/pacbio/PacbioPoolAliquotEdit.vue
seenanair: Sorry, that was part of a trial, will remove it

BenTopping (dismissed):

stevieing (dismissed):
Great. I am kind of hoping that we are moving to using more types and we can start standardising them.

seenanair:
@BenTopping @stevieing  There is a minor modification where the tailwind classes bit  is changed . Please review.

BenTopping (approved):

stevieing (approved):

PR #1695, opened 2024-05-10
Y24-015 - Cypress uses a different traction-service base URL

### LINKED ISSUE #1575: Y24-015 As a developer I would like cypress to use a different traction-service URL so we don't get false passing tests locally
**Describe the Housekeeping**
As a developer I would like cypress to use a different traction-service URL so we don't get false passing tests locally. Currently the e2e tests point to the default traction-service address and port. This means if a test is not stubbing a request properly it could make it through to the locally running traction-service instance and return data.
This makes requests not being stubbed harder to identify and can cause tests to pass locally but not on CI (where the request is not being stubbed and not reaching a traction-service instance)

**Developer knowledege**
Seena N, Ben T

**Acceptance Criteria**

- [ ] Cypress e2e tests go to a non default traction-service url/port. E.g. localhost:3002
- [ ] Cypress stubs are updated to point to the new traction-service url

**Additional context**

SHIV5T3R:
Closes #1575 

- The commands for cypress e2e tests will have the `VITE_TRACTION_BASE_URL` env variable overridden to "http://localhost:3004" to avoid conflicting with local instances of the traction-service.


package.json
```diff
@@ -10,10 +10,10 @@
     "build:js": "vite build",
     "test:unit": "vitest run",
     "test:e2e": "test:e2e:all",
-    "test:e2e:all": "npm-run-all -p start test:cypress:all",
-    "test:e2e:one": "npm-run-all -p start test:cypress:one",
-    "test:cypress:all": "cypress run",
-    "test:cypress:one": "cypress open",
+    "test:e2e:all": "VITE_TRACTION_BASE_URL=http://localhost:3004 npm-run-all -p start test:cypress:all",
```
BenTopping, line 13: Line 13 and 14 are aliases for 15 and 16 so the prefix should only be needed on 15 & 16.

BenTopping: My mistake - 'test:e2e:all' and 'test:e2e:one' actually call/alias to two commands so you need the override prefix there because otherwise the 'start' command wont pick up the override.

BenTopping: I suspect this is why it appeared as though cypress wasn't picking up the env variables because 'start' doesn't pick them up

PR #1727, opened 2024-05-29
UAT feedback updates for DPL-1077

seenanair:
- Volume field initialised to available_volume instead of original volume
- Differently coloured badges for 'available_volume" and 'original_volume" information displayed

### DIFF
--- a/src/components/pacbio/PacbioLibraryForm.vue
+++ b/src/components/pacbio/PacbioLibraryForm.vue
@@ -34,7 +34,7 @@
               v-if="formLibrary.used_volume != null"
               :tooltip-text="'Used volume is ' + formLibrary.used_volume"
             >
-              <traction-badge id="library-used-volume" colour="sanger-yellow"
+              <traction-badge id="library-used-volume" colour="sanger-pink"
                 ><TractionInfoIcon class="mr-2" />{{ formLibrary.used_volume }}</traction-badge
               >
             </traction-tooltip>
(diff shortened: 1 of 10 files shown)


stevieing (changes requested):
Unfortunately it does not seem to be working correctly unless I am missing something:
- I can enter a volume greater than the available volume
- I can also create another run with the same library with any volume

I also have some structural comments below:
- There are quite a few things going on here. It all works but has got quite complex. You are using a type but a lot of the heavy lifting is being done outside of that type.
- I have also noticed that you are using ref for things that are purely object i.e. well and localUsedAliquots. Maybe it would be an idea to use reactive? That way you do not need to unwrap it every time. As long as you are not changing the whole object it should work.

seenanair:
Regarding the errors :-
I rechecked these and both these cases seems to be working in mine. Just to make sure is "dpl_1076_check_library_volume_in_runs" feature flag turned on in yours? 
If so, it would be good to have a call regarding this, whenever you are free.

Regarding the structural comments, I agree, but all of these were already there and is not part of the changes I made as part of this PR. I suspect it can lead to major refactoring changes which I am not sure is the right thing to do at this stage. Please share your comments.

> Unfortunately it does not seem to be working correctly unless I am missing something:
> 
> * I can enter a volume greater than the available volume
> * I can also create another run with the same library with any volume

> I also have some structural comments below:
> 
> * There are quite a few things going on here. It all works but has got quite complex. You are using a type but a lot of the heavy lifting is being done outside of that type.
> * I have also noticed that you are using ref for things that are purely object i.e. well and localUsedAliquots. Maybe it would be an idea to use reactive? That way you do not need to unwrap it every time. As long as you are not changing the whole object it should work.

stevieing (approved):
Great. Working as expected.

We can get it onto UAT but I would like to discuss the structure to ensure it is maintainable.

PR #1732, opened 2024-05-31
Y24-085 libraries with same source

seenanair:
This PR includes fix for the issue - 'PacBio pools/new page does not allow for multiple libraries from the same source. When you scan in multiple libraries from the same source they select/deselect together.'

-  used_aliquots will be mapped against source_id instead of request id
-  pacbioPoolCreate store state,getters and actions updated with source_id
-  All Pacbio pool components will be using source_id

This PR also includes refactoring on pacbio run creation which includes - initialisation of the volume with available_volume in used_aliquot type , PacbioRunWellEdit component optimisation


src/components/labware/PacbioRunWell.vue
```diff
@@ -196,15 +196,24 @@ const updateUsedAliquotSource = async (barcode) => {
...
+    {
+      source_id: id,
+      source_type,
+      concentration,
+      insert_size,
+      template_prep_kit_box_barcode,
+      barcode,
+      volume,
+    },
+    source_type === 'Pacbio::Library'
```
BenTopping, line 209: Could we do this check above and then set the volume and available_volume based on it? something like.

```
const available_volume = source_type === 'Pacbio::Library' ? getAvailableVolume.. : null

const used_aliquot = createUsedAliquot(
{
...
volume: available_volume || volume,
available_volume
}
```

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -128,9 +128,9 @@ defineProps({
 
 // Define refs for the component
 // well object
-const well = ref({})
+const well = reactive({ used_aliquots: [], loading_target_p1_plus_p2: '' })
```
BenTopping, line 131: loading_target_p1_plus_p2 is SMRTLink version specific so it feels like it shouldn't be here.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -208,14 +208,14 @@ const formatLoadingTargetValue = (val) => {
     if (decimalPercentageRegex.test(val)) {
       return val
     } else {
-      return isNaN(well.value.loading_target_p1_plus_p2) ? 0 : well.value.loading_target_p1_plus_p2
+      return isNaN(well.value.loading_target_p1_plus_p2) ? 0 : well.loading_target_p1_plus_p2
```
BenTopping, line 211: is well.value needed?

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -283,27 +278,26 @@ const updateUsedAliquotSource = async (row, barcode) => {
...
+        id: row.item.id || '',
+        source_id: tubeContent.id,
+        source_type: type,
+        barcode,
+        volume: tubeContent.volume,
+        concentration: tubeContent.concentration,
+        insert_size: tubeContent.insert_size,
+        template_prep_kit_box_barcode: tubeContent.template_prep_kit_box_barcode,
+      },
+      type === 'Pacbio::Library'
```
BenTopping, line 292: Similar comment to PacbioRunWell comment, feels like this logic should be above.

src/stores/pacbioPoolCreate.js
```diff
@@ -246,29 +249,27 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
       const requests = state.resources.requests
-      const selectedRequests = state.used_aliquots
+      const selectedUsedAliquots = state.used_aliquots
       let val = []
-      if (ids) {
-        val = ids.map((id) => {
-          return { ...requests[id], selected: !!selectedRequests[`_${id}`] }
+      if (source_obj) {
+        const source_id = source_obj.source_id ?? source_obj.id
+        const used_aliquot = state.used_aliquots[`_${source_id}`]
```
BenTopping, line 265: Feels like a sensible time to we get rid of the '_' in the used_aliquots key

src/stores/pacbioPoolCreate.js
```diff
@@ -246,29 +249,27 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
+        const selected = !!used_aliquot
+        val = source_obj['requests']?.map((id) => {
+          return { ...requests[id], selected, source_id }
         })
       } else {
         val = Object.values(requests).map((request) => {
-          return { ...request, selected: !!selectedRequests[`_${request.id}`] }
+          return { ...request, selected: !!selectedUsedAliquots[`_${request.source}`] }
```
BenTopping, line 272: Is this suppose to be `!!selectedUsedAlqiuots['_${request.source_id}']`?

src/stores/utilities/usedAliquot.js
```diff
@@ -31,6 +36,8 @@ function createUsedAliquot(attributes) {
     ...defaultUsedAliquotAttributes,
     ...attributes,
     tag_id: (attributes && (attributes.tag || attributes.tag_id)) ?? null,
+    volume: initialiseToAvailableVolume ? availableVolume : formatValue(attributes?.volume),
```
BenTopping, line 39: Discussed in previous comments, I think we should pass in available_volume and volume in attributes and set the volume upfront instead of mutating the values in here.

src/views/pacbio/PacbioPoolCreate.vue
```diff
@@ -160,21 +160,21 @@ const search = async (value) => {
...
   if (!labware) {
     labware = pacbioPoolCreateStore.selectedPlates.find((plate) =>
       pacbioPoolCreateStore
         .wellList(plate.wells || [])
-        .some((well) => well.requests.includes(request.id)),
+        .some((well) => well.id === aliquot.source_id),
     )
   }
-  aliquotSelectionHighlightLabware.value = { labware, request }
+  aliquotSelectionHighlightLabware.value = { labware, request: aliquot }
```
BenTopping, line 178: can we rename the request here to aliquot so its just ` ...Labware.value = { labware, aliquot }`

BenTopping (commented):
Quite a lot in here so apologies if i have misunderstood some bits.

src/components/labware/PacbioRunWell.vue
seenanair: The idea was to contain it within the used_aliquot type, so we don't need to set this in different places around that which will help any modification in future.
But, I think we can use this source_type === 'Pacbio::Library' check also within the type itself, thereby avoiding these checks to be done everytime. Hope it make sense. Please let me know otherwise.

src/components/pacbio/PacbioRunWellEdit.vue
seenanair: Good spot, It is not needed as it is reactive now.

src/stores/pacbioPoolCreate.js
seenanair: I thought about it but didn't do as t is already quite a lot refactoring and changes every where.

src/stores/pacbioPoolCreate.js
seenanair: I think it is kept as 'source' inside the request object.I will double check it.

src/stores/pacbioPoolCreate.js
BenTopping: Sure, might be worth making an issue then so we can make the technical debt visible.

src/components/labware/PacbioRunWell.vue
BenTopping: OK, yeah I think the logic should be in one place.

src/stores/pacbioPoolCreate.js
seenanair: Sure, will do. 👍

src/views/pacbio/PacbioPoolCreate.vue
seenanair: Done

src/stores/pacbioPoolCreate.js
seenanair: It is returned as 'well' or 'tube'  inside request from service.

src/components/pacbio/PacbioLabwareSelectedList.vue
```diff
@@ -143,25 +143,29 @@ const pacbioPoolCreateStore = usePacbioPoolCreateStore() // A composable store f
...
-        .wellList(plate.wells)
-        .flatMap((well) => pacbioPoolCreateStore.requestList(well.requests || []))
-    } else {
-      const tube = pacbioPoolCreateStore.selectedTubes.find(
-        (item) => item.barcode === labware.barcode,
-      )
-      return pacbioPoolCreateStore.requestList(tube.requests || [])
-    }
-  })
+  const requests = props.labware
```
stevieing, line 156: Advisory Not needed for this cycle: Again I think there is oppurtunity for refactor here. There are a lot of similarities between the two paths so you could reduce it into a method with function passing.

stevieing (changes requested):
Just a couple of minor things and some advisory. There were other things that we could also do but we can delay for now. Maybe it would be worth creating a refactoring story with anything we want to do?

stevieing (approved):

stevieing (commented):
Well done. Still think there are further improvements we could do but more than enough for now.

PR #1748, opened 2024-06-11
DPL-1041 - As a stakeholder james w I would traction to track Pacbio Smrtlink library types

stevieing:
Closes #

- Refactor PacbioRunWellComponents.json and run.js to add library_type
- Added default well atrributes to the json file so it is all in one place
- Renamed PacbioRunWellComponents to PacbioRunWellSmrtLinkOptions as it is more in keeping with what it is doing
- Added .json file suffix


src/stores/pacbioRunCreate.js
```diff
@@ -466,7 +460,7 @@ export const usePacbioRunCreateStore = defineStore('pacbioRunCreate', {
      * Sets the defaultWellAttributes
      */
     setDefaultWellAttributes() {
-      this.defaultWellAttributes = defaultWellAttributes()
+      this.defaultWellAttributes = PacbioRunWellSmrtLinkOptions.defaultAttributes
```
BenTopping, line 463: This will need to be spread `{...PacbioRunWellSmrtLinkOptions.defaultAttributes }` as we don't want to mutate the original object as we need to reset them to their original state when the run is reset.

BenTopping (changes requested):
Looks good, just a small issue with the values not being reset when the run reset is pressed.

BenTopping (approved):

PR #1756, opened 2024-06-14
DPL-1093 - As developers we want to find out how to modify the e2e fixtures so we can re-use the factory wrapper that we are creating so we do not have to duplicate requests

stevieing:
Closes #

- Created a Factory that does not import data from a file as this does not work in e2e tests.
- Created a BaseFactory that includes content and an Axios and Fetch response
- Replaced fixtures in e2e tests with factory
- 


dasunpubudumal (dismissed):

tests/unit/components/labelPrinting/LabelPrintingForm.spec.js
```diff
@@ -24,12 +19,14 @@ const evt = {
...
-const printerRequestFactory = RequestFactory('printers', false)
+const printersFactory = PrinterFactory()
 
 const plugins = [
   ({ store }) => {
     if (store.$id === 'root') {
-      store.api.v2.traction.printers.get = vi.fn().mockResolvedValue(printerRequestFactory.response)
+      store.api.v2.traction.printers.get = vi
+        .fn()
+        .mockResolvedValue(printersFactory.responses.fetch)
```
seenanair, line 29: I hope once we move everything to fetch, it will be just be printersFactory.responses.

tests/factories/BaseFactory.js
```diff
@@ -0,0 +1,27 @@
...
+      },
+      fetch: {
+        status: 200,
+        statusText: 'OK',
+        ok: true,
+        json: () => Promise.resolve({ ...data }),
+      },
+    },
+  }
+}
```
seenanair, line 25: Nice abstraction 👍

tests/factories/PacbioSmrtLinkVersionFactory.js
```diff
@@ -0,0 +1,100 @@
...
+            },
+          },
+        },
+      },
+    ],
+  }
+
+  const storeData = dataToObjectById({ ...data })
+
+  return { ...BaseFactory(data), storeData }
```
seenanair, line 97: I hope keeping store data will simplify unit tests a lot, as we replicate much of it across tests.

seenanair (approved):

stevieing:
@seenanair and @dasunpubudumal some stuff was merged into develop this morning so I have had to fix the conflicts.

dasunpubudumal (approved):

tests/unit/components/labelPrinting/LabelPrintingForm.spec.js
stevieing: Maybe. I like the idea of it being descriptive and it gives us the flexibility of using different response types. I think we might need it if we do location tracking as Labwhere does not use JSON API resource.

tests/factories/PacbioSmrtLinkVersionFactory.js
stevieing: Yes. I decide to add this at this level as each factory will create store data differently. We can see how it progresses ...

seenanair: Yes, it is best to keep it here because it varies a lot between stores and probably needs many steps for complicated stores. Maybe after we do all these, we might find some patterns and then we can do further abstraction.

PR #1757, opened 2024-06-14
Have consistent ordering between pipeline pages

StephenHulme:
Have consistent ordering between pipeline pages

- Leave the button order unchanged
- Add comment to specify that the order is intentional and based on user workflow
- Increase robustness of button tests

### DIFF
--- a/tests/unit/views/TractionDashboard.spec.js
+++ b/tests/unit/views/TractionDashboard.spec.js
@@ -48,6 +48,12 @@ describe('TractionDashboard.vue', () => {
     })
 
     describe('route buttons', () => {
+      it('will have buttons sorted by workflow', () => {
+        const buttons = box.findAll('a')
+        const buttonNames = buttons.map((button) => button.text())
+        expect(buttonNames).toEqual(['Samples', 'Libraries', 'Runs'])
+      })
+
       it('will have a samples button', async () => {
         const button = findByText(box, 'Samples')
         button.trigger('click')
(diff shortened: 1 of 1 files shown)


BenTopping:
I don't think there is a problem with the changes just have two comments:

- The lab might be used to the current layout so reordering it might not be helpful for them.
- I thought of the current order as a logical progression through the pipeline (samples -> libs/pools -> runs), not sure if it was intended this way but something to consider.

seenanair:
It looks fine, but I agree with Ben's comments.

I suppose the current order is maintained according to the flow of the pipeline process, which is a good thing to keep. 

Also, since lab users are accustomed to this layout, there is a risk in how they perceive any changes. It might be beneficial to present a mock-up to them before proceeding.

Again just a thought as you are working on this, it would be good to explore ways to avoid using tabs, which will provide more screen space for the pages. Instead, consider utilizing already existing menus. For example, the menu bar could include the pipeline with each of these tabs as a sub-menu item.

stevieing:
I think I would tend to agree with @seenanair and @BenTopping . The order is logical as it follows the user workflow. In cases like this it is better if changes like this are user led. I do think it is an idea to revisit styling and UX regularly but we will probably be doing this again in some bigger pieces of work.

stevieing:
> Thanks for the helpful feedback all! I didn't realise that it was based on workflow order (although I do feel, that pools and pools/new should be together). I'll leave it as it is and add a small comment in the code making the intended order clearer. (I updated some tests for robustness as well, I'll keep them in too).

The test updates were fine. Pool/new was created before pools so we left it like that. It also means 1 less click for users as they create pools by scanning so it is not necessary to go to the pools page.

I am sure we will review the UI when the next piece of bug work comes along which is automation which may transform the UI.

stevieing (approved):

PR #1765, opened 2024-06-18
add active flag in pipeline config file

wendyyang:
Closes #

add active flag in pipeline config file
use the active flag to filter pipelines which are active to be display.
- 


seenanair (commented):
Looks fine. I think the unit tests need to be updated for testing 'active' flag.

wendyyang:
> Looks fine. I think the unit tests need to be updated for testing 'active' flag.

Thanks Seena! i have updated the unit test to include 'active' flag. Could you please have a second code review?

seenanair (dismissed):

stevieing (changes requested):
I think you need to add a test for when one of the pipelines is turned off. It may mean that you need to pass the pipelinesConfig in in such a way that you can modify it in the test. Hope that makes sense?

src/views/TractionDashboard.vue
```diff
@@ -31,7 +31,7 @@ export default {
     TractionLink,
   },
   computed: {
-    pipelines: () => PipelinesConfig,
+    pipelines: () => PipelinesConfig.filter((pipeline) => pipeline.active == true),
```
stevieing, line 34: You can do `PipelinesConfig.filter(({active}) => active)`

wendyyang:
> I think you need to add a test for when one of the pipelines is turned off. It may mean that you need to pass the pipelinesConfig in in such a way that you can modify it in the test. Hope that makes sense?

updated TractionDashboard.spec.js to be able to pass test when some of the pipelines to be turned off.

tests/unit/views/TractionDashboard.spec.js
```diff
@@ -1,180 +1,181 @@
...
-    })
-
-    describe('route buttons', () => {
-      it('will have a samples button', async () => {
-        const button = box.findAll('a')[0]
-        expect(button.text()).toEqual('Samples')
-        button.trigger('click')
-        await flushPromises()
-        expect(wrapper.vm.$route.path).toBe('/saphyr/samples')
+  if (active_pipeline_names.includes('saphyr')) {
```
seenanair, line 32: Conditionally running tests may not be ideal as it can cause non-deterministic tests and can be confusing. It would be better to use a mocking setup, which allows you to mock the file contents so we can control the testing data.

tests/unit/views/TractionDashboard.spec.js
```diff
@@ -1,180 +1,181 @@
...
+  }
 
-    describe('route buttons', () => {
-      it('will have a plates button', async () => {
-        const platesButton = box.findAll('a')[0]
-        expect(platesButton.text()).toEqual('Plates')
-        platesButton.trigger('click')
-        await flushPromises()
-        expect(wrapper.vm.$route.path).toBe('/pacbio/plates')
+  if (active_pipeline_names.includes('pacbio')) {
```
seenanair, line 74: Same comment as above for conditional testing.

tests/unit/views/TractionDashboard.spec.js
wendyyang: Hi Seena, 
Thanks for your comments. I made a mock config file for testing(I push to git already) but I found it still not able to pass the test as the dashboard pipelines are read from real config file. so comparing dashboard pipeline and mock pipeline config file won't work. Any possible that the dashboard pipeline are from mock config file when testing?

seenanair: Hi Wendy, I think we don’t need to create a mock file instead we can mock what the file content is, for example , some thing like this
vi.mock(‘@/config/PipelinesConfig’, () => (
{
    “name”: “pipeline”,
    “title”: “Pipileline title”,
    “description”: “Testing”,
    “routes”: [“samples”, “libraries”, “runs”],
     active : false
  },
 ))

I am happy to catchup tomorrow, possibly in the afternoon, if you would like to discuss further

wendyyang: Hi Seena,
Thanks for your suggestion. I made further changes for the test. Please review again.

seenanair (commented):
Looking good 👍  Happy to approve once you resolve the conflicts.

wendyyang:
> Looking good 👍 Happy to approve once you resolve the conflicts.

Thank you! Seena. I have now resolved the conflicts.

seenanair (approved):

PR #1806, opened 2024-07-05
Refactor pacbio/requests

dasunpubudumal:
Closes #

- Create factories and update the tests to use fetch for pacbio requests.


src/store/traction/pacbio/requests/actions.js
```diff
@@ -1,11 +1,15 @@
-import handleResponse from '@/api/v1/ResponseHelper'
+import handleResponse from '@/api/v2/ResponseHelper'
```
seenanair, line 1: Just a thought, Better to include file extension which speeds up vite build process.

tests/factories/TractionPacbioSamplesFactory.js
```diff
@@ -0,0 +1,208 @@
...
+      ],
+    },
+    status: 200,
+    statusText: 'Success',
+  }
+
+  return BaseFactory(data)
+}
+
+export default TractionPacbioSamplesFactory
```
seenanair, line 208: I believe that for all other existing factories, we start with the pipeline name. Would it be better to use `PacbioSamplesFactory` instead of `TractionPacbioSamplesFactory` to maintain uniformity and follow the same naming convention?

tests/unit/store/traction/pacbio/requests/actions.spec.js
```diff
@@ -1,63 +1,28 @@
...
   })
 
   describe('setRequests', () => {
     it('fetches the requets from the service, and commits them', async () => {
       const commit = vi.fn()
       const get = vi.fn()
       const getters = { requestsRequest: { get: get } }
-      const requests = Data.TractionPacbioSamples
-      requests.data.data.splice(2, 11)
+      const requests = tractionPacbioSamplesFactory.responses.fetch
```
seenanair, line 20: May be this assignment line can be avoided and use it directly with mockReturnValue.

seenanair (commented):
Nice refactoring. Just shared some minor suggestions.

seenanair:
I couldn't see any old data or fixture data files removed. It would be good to check that as well.

dasunpubudumal:
> I couldn't see any old data or fixture data files removed. It would be good to check that as well.

Nice catch. I actually forgot to remove it.

stevieing (changes requested):
I concur with Seena. Removing any old files and using the same naming convention for factories.

stevieing (approved):

PR #1821, opened 2024-07-17
Y24-138 Moving Pacbio plates store from axios to fetch

seenanair:
Closes #

- Create a PacbioPlateRequests factory 
- Refactor unit and e2e tests to use the PacbioPlateRequests factory
- Change to v2 (fetch) in store getters
- Remove PacbioPlateRequests fixture data file 
- Remove PacbioPlateRequests data file


src/api/Config.js
```diff
@@ -155,6 +155,125 @@ const config = [
       },
     ],
   },
+  {
+    name: 'tractionv2',
```
stevieing, line 159: What is this for? We do not have a traction api v2 in the service.

src/store/traction/pacbio/plates/actions.js
```diff
@@ -1,12 +1,16 @@
-import handleResponse from '@/api/v1/ResponseHelper'
+import handleResponse from '@/api/v2/ResponseHelper'
 import { groupIncludedByResource } from '@/api/JsonApi'
 
-const setPlates = async ({ commit, getters }, filter, page) => {
+const setPlates = async ({ commit, getters }, options) => {
```
stevieing, line 4: Does this make things better? It feels less explicit. Did we not create a type to pass into requests?

src/store/traction/pacbio/plates/actions.js
```diff
@@ -1,12 +1,16 @@
...
   const request = getters.getPlates
-  const promise = request.get({ page, filter })
+  const promise = request.get({ page: options?.page, filter: options?.filter })
```
stevieing, line 6: You can just destructure options as in `request.get({ ...options })`

src/store/traction/pacbio/plates/actions.js
```diff
@@ -18,10 +22,10 @@ const findPlate = async ({ getters }, filter) => {
   const request = getters.getPlates
   const promise = request.get({ filter, include: 'wells.requests' })
   const response = await handleResponse(promise)
-  const { success, data: { data, included = [] } = {} } = response
+  const { data, included = [] } = response.body
```
stevieing, line 25: You can do `const { success, body: { data, included = [] } } = response`

tests/e2e/specs/pacbio/pacbio_plates_view.cy.js
```diff
@@ -1,29 +1,23 @@
...
-          },
-        ],
-      },
-    })
     // When we type 1 into input per page we search for page size 1
-    cy.intercept('/v1/pacbio/plates?page[size]=1&page[number]=1', {
-      fixture: 'pacbioPlatesRequest.json',
+    cy.wrap(PacbioPlatesRequestFactory()).as('pacbioPlateRequestFactory')
+    cy.get('@pacbioPlateRequestFactory').then((pacbioPlateRequestFactory) => {
+      cy.intercept('GET', '/v1/pacbio/plates?page[size]=1&page[number]=1', {
```
stevieing, line 11: Are you doing an intercept twice here?

tests/unit/components/pacbio/PacbioLabwareSelectedList.spec.js
```diff
@@ -37,19 +38,20 @@ function mountWithStore({ state = {}, stubActions = false, plugins = [], props }
   const storeObj = usePacbioPoolCreateStore()
   return { wrapperObj, storeObj }
 }
+const pacbioPlatesRequestFactory = PacbioPlatesRequestFactory()
 
 describe('PacbioLabwareSelectedList', () => {
   let wrapper, store
   const plates = dataToObjectById({
-    data: Data.PacbioPlatesRequest.data.data,
+    data: pacbioPlatesRequestFactory.content.data,
```
stevieing, line 46: You could possibly move this to a storeData function?

tests/unit/store/traction/pacbio/plates/mutations.spec.js
```diff
@@ -1,29 +1,17 @@
...
-    ]
-  })
-
   it('"setPlates" sets "state.plates" to the given plates', () => {
     const state = {
       plates: {},
     }
-    mutations.setPlates(state, plates)
-    expect(state.plates).toEqual(dataToObjectById({ data: plates, includeRelationships: true }))
+    mutations.setPlates(state, pacbioPlatesRequestFactory.storeData)
```
stevieing, line 12: You can also move this function to the storeData function. I just added a selected which pulls out a small number of records.

tests/unit/views/pacbio/PacbioPlateIndex.spec.js
```diff
@@ -34,21 +35,23 @@ describe('PacbioPlates.vue', () => {
...
     })
   })
 
   describe('Plate display button', () => {
     let button
 
     it('is present for each plate', () => {
-      // btn-61 because that is the first plate ID in the mocked data
-      button = wrapper.find('#details-btn-61')
+      const id = pacbioPlatesRequestFactory.content.data[0].id
```
stevieing, line 46: Again you can move this to the storeData.

tests/unit/views/pacbio/PacbioPoolCreate.spec.js
```diff
@@ -77,7 +79,7 @@ describe('PacbioPoolCreate', () => {
   })
 
   const wells = dataToObjectById({
-    data: Data.PacbioPlatesRequest.data.included.slice(0, 4),
+    data: pacbioPlatesRequestFactory.content.included.slice(0, 4),
```
stevieing, line 82: If you move this to the storeData you can reuse what you had in the tests/unit/components/pacbio/PacbioLabwareSelectedList.spec.js

stevieing (changes requested):
Nice. A couple of things I didn't understand and a few things that you could move to the storeData to tidy things up more.

tests/unit/store/traction/pacbio/plates/actions.spec.js
```diff
@@ -1,98 +1,57 @@
...
-          },
-        ],
-      }
+      get.mockReturnValue(successResponse)
+
       // Barcode provided is first plate in Data.PacbioPlatesRequest
       const plate = await Actions.findPlate({ commit, getters }, { barcode: 'DN814327C' })
 
-      expect(plate).toEqual(expectedPlate)
+      expect(plate).toEqual(pacbioPlatesRequestFactory.expectedPlates)
```
stevieing, line 50: I did something similar but used the storeData and used storeData.selected etc. Might be worth following a similar model.

src/api/Config.js
seenanair: That's something on trial, that I forgot to remove 🤦‍♀️

src/store/traction/pacbio/plates/actions.js
seenanair: `setPlates` is called using composable function `fetchWithQueryParams`  and it should be an object because that's how fetchWithQueryParams is wrapping parameters which is as follows

` const page = { size: page_size.value.toString(), number: page_number.value.toString() }
    const filter = buildFilter(filterOptions)
    const { success, errors, meta } = await fetcher({ page, filter })`

I didn't want to change the composable function since it might be used in many places. 

The parameter name changed to 'filterOptions' to make it more clear.

src/store/traction/pacbio/plates/actions.js
seenanair: Good spot 👍

tests/e2e/specs/pacbio/pacbio_plates_view.cy.js
seenanair: Yes, intercepting twice was required because of different page size query parameters. Initially, when a page is first loaded, the size is 25 and we are changing the page size in test to 1. I have now replaced it with a RegExp so that it can work in both cases.

tests/unit/store/traction/pacbio/plates/actions.spec.js
seenanair: This actually requires a method to do the same kind of data restructuring as it does in actions to reformat the plate requests to well. So, i don't know how useful it will be because we end up replicating similar code that we intend to test.

stevieing (dismissed):

stevieing (approved):

PR #1834, opened 2024-07-24
Y24-155 - Pool volume check on PacBio run

### LINKED ISSUE #1785: Y24-155 - As a user (James W) I would like there to be a volume check when adding a pool to a PacBio run.
**User story**
As a user (James W) I would like there to be a volume check when adding a pool to a PacBio run.

James W, Ben T, Seena N

James W

**Acceptance criteria**
- [ ] The available volume for a pool should be visible when editing/creating a well in a run. This should behave the same as the PacBio library available volume label. This should be feature flagged.

**Dependencies**
- [Y24-153](https://github.com/sanger/traction-service/issues/1348) Dependent as this story exposes the necessary data and behaviour for volume checks.

**Additional context**
We check whether the tube is a pool or library in several places in the PacBio run to facilitate correct volume checking, but now they both require volume checking a lot of these can be removed.

BenTopping:
Closes #1785

- updates getAvailableVolumeForLibraryAliquot to work with both pools and libraries


seenanair (approved):
Looking good 👍

seenanair:
@BenTopping Just to let you know, I didn't do any testing.

stevieing (changes requested):

src/stores/pacbioRunCreate.js
```diff
@@ -520,53 +519,57 @@ export const usePacbioRunCreateStore = defineStore('pacbioRunCreate', {
...
+      const available_volume = sourceStore[sourceId]?.available_volume || 0
+
+      // Calculate the sum of the volume of all the new aliquots used in wells that are from the source
+      let used_aliquots_volume = 0
       Object.values(this.wells).forEach((plate) => {
         Object.values(plate).forEach((well) => {
           well.used_aliquots?.forEach((aliquot) => {
-            // Existing aliquots should not be counted as they are already taken into account in the library available volume
+            // Existing aliquots should not be counted as they are already taken into account in the source available volume
             // This has the issue that if an existing aliquots volume is changed it will not be reflected in the available volume
```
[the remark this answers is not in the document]
BenTopping: I can go over this logic with you at some point if that would be useful. This loop is used to identify how much extra volume has been used on top of the libraries original available volume.

src/components/pacbio/PacbioRunWellEdit.vue
```diff
@@ -291,6 +303,9 @@ const updateUsedAliquotSource = async (row, barcode) => {
  * This function is called when the modal is shown for a specific position and plate number.
  */
 const setupWell = async () => {
+  pool_volume_validation_flag.value = await checkFeatureFlag(
+    'y24_155_pacbio_run_pool_volume_validation',
+  )
```
seenanair, line 308: Just a thought. Do you think this way of using a feature flag is better? The only thought I have is that the other way is more confined, as it restricts the presence of the feature flag to one place, while this approach spreads it across many places.

BenTopping: The other way being using the feature flag component? I did it this way because the changes aren't to display different components depending on the flag but to a hide component based on the **flag and type of aliquot.**

seenanair (approved):

PR #1849, opened 2024-08-01
Y24-200 - Run volume tracking bug

### LINKED ISSUE #1819: Y24-200 - [BUG] - Unable to add a library several times to the same run.
**Describe the bug**
Hi, I encountered a bug this week setting up a run. I had a Sequel II run that had the same libraries being run multiple times. Due to the volume tracking fields I was unable to set the run up in one go as the available volume for each library was being deducted with the first instance on the run even if edited to a the volume used for each well being sequenced.
The run in question is TRACTION-RUN-1377.
I managed to work around this issue by creating the run with each library appearing once and then editing each run to add the duplicate library and then editing again for the triplicate. It may be rare that his functionality would be used on the Revio but it should be possible to sequence the sample library/pool more than once on the same run. Maybe even being able to override once the volume has hit zero. At this point in the process we will physically have a library to load so should not be prevented from loading if library volume is 0 (but this is the best point currently to register volume of library used).
In this particular instance I have used the entire library that was available and then split out the resulting reaction mix across 2 or more cells to use up the remaining library in the most efficient way possible. So, I would be happy to set the volume correctly on the first cell for the library but for any additional cells from the same library would not need further updating.
Next week I will also be loading some of the same libraries, at this moment in time I have just estimated how much library I have used for each cell so that I don't get a zero volume error when I try to create a run. However If we decided to not proceed any further the remaining library volumes would be incorrect (without further editing).

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**

BenTopping:
Closes #1819 

- Updates the way available_volume is calculated for Libraries and Pools in a PacBio run
- Updates the way available_volume is shown in PacbioRunWellEdit.
- Updates how available_volume and volume are calculated in the createdUsedAliquot function.

### DIFF
--- a/src/components/labware/PacbioRunWell.vue
+++ b/src/components/labware/PacbioRunWell.vue
@@ -196,17 +196,20 @@ const updateUsedAliquotSource = async (barcode) => {
   const tubeContent = store.tubeContentByBarcode(barcode)
   // Determine the source type for the used aliquot
   const source_type = tubeContent.type === 'pools' ? 'Pacbio::Pool' : 'Pacbio::Library'
+  const available_volume = store.getAvailableVolumeForAliquot({
+    sourceId: tubeContent.id,
+    sourceType: source_type,
+    volume: 0,
+  })
   /*id set to null because we are creating a new used aliquot. This is very important for calculating 
   the available volume where it checks for existing aliquots using id*/
-  const used_aliquot = createUsedAliquot(
-    { ...tubeContent, source_id: tubeContent.id, id: null },
-    () =>
-      store.getAvailableVolumeForAliquot({
-        sourceId: tubeContent.id,
-        sourceType: source_type,
-        volume: 0,
-      }),
-  )
+  const used_aliquot = createUsedAliquot({
+    ...tubeContent,
+    source_id: tubeContent.id,
+    id: null,
+    volume: available_volume,
+    available_volume,
+  })
   well.used_aliquots.push(used_aliquot)
   store.updateWell({ well, plateNumber: props.plateNumber })
 }
--- a/src/components/labware/PacbioRunWell.vue
+++ b/src/components/labware/PacbioRunWell.vue
@@ -196,13 +196,14 @@ const updateUsedAliquotSource = async (barcode) => {
   const tubeContent = store.tubeContentByBarcode(barcode)
   // Determine the source type for the used aliquot
   const source_type = tubeContent.type === 'pools' ? 'Pacbio::Pool' : 'Pacbio::Library'
+  // We get the full available volume from the given source by passing a volume of 0
+  // This is used to calculate the initial volume of the used aliquot
   const available_volume = store.getAvailableVolumeForAliquot({
     sourceId: tubeContent.id,
     sourceType: source_type,
     volume: 0,
   })
-  /*id set to null because we are creating a new used aliquot. This is very important for calculating 
-  the available volume where it checks for existing aliquots using id*/
+  /*id set to null because we are creating a new used aliquot */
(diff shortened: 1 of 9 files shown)


SHIV5T3R (approved):

src/components/labware/PacbioRunWell.vue
```diff
@@ -196,17 +196,20 @@ const updateUsedAliquotSource = async (barcode) => {
...
-        volume: 0,
-      }),
-  )
+  const used_aliquot = createUsedAliquot({
+    ...tubeContent,
+    source_id: tubeContent.id,
+    id: null,
+    volume: available_volume,
+    available_volume,
+  })
```
seenanair, line 212: Just a thought: now that we are dynamically calculating available_volume every time, would it be better to refactor the createUsedAliquot function to handle the calculation of available_volume internally so we don't need to repeat two calls? This would make the code cleaner.

src/stores/pacbioRunCreate.js
```diff
@@ -527,55 +518,52 @@ export const usePacbioRunCreateStore = defineStore('pacbioRunCreate', {
...
-        // If its a new aliquot we need to add the volume back in to the total available volume
-        // because it was removed as part of the library_used_aliquots volume but it should be available in this instance of aliquot
-        total_available_volume = parseFloat(total_available_volume) + parseFloat(volume)
-      }
+      let total_available_volume = (
+        available_volume +
+        original_aliquot_volume -
+        used_aliquots_volume +
+        parseFloat(volume)
+      ).toFixed(2)
```
seenanair, line 563: Would it be better to add a comment to describe a little more as this bit complex and confusing for someone looking at it in future.

tests/unit/stores/utilities/pool.spec.js
```diff
@@ -28,11 +30,7 @@ describe('pool', () => {
       expect(usedAliquot.volume).toBe(10)
       expect(usedAliquot.concentration).toBe(5)
       expect(usedAliquot.insert_size).toBe(1)
-    })
-    it('correctly sets the default volume when initialiseToAvailableVolume is provided', () => {
-      const initialiseToAvailableVolume = () => '20'
-      const usedAliquot = createUsedAliquot({ volume: 10 }, initialiseToAvailableVolume)
-      expect(usedAliquot.volume).toBe(20)
+      expect(usedAliquot.available_volume).toBe(10)
     })
```
seenanair, line 34: Is there a test already exists either in unit or e2e to check the context where we are adding the same library twice in a run, to ensure that this fix is covered? If not, would it be better to add one?

seenanair (commented):
Great fix. 👍 
Just added a few thoughts.
Also, I haven't tested it locally.

src/components/labware/PacbioRunWell.vue
BenTopping: available_volume requires the run store state so we need to pass the whole state into the createUsedAliquot function. I am also not sure available_volume should live in the aliquot as its actually a mixture of the aliquot data and library/pool data but couldn't find a better place to put it.

seenanair: Ok, hmm, that makes sense. yes, it looks like getAvailableVolumeForAliquot needs lots of store state variables.

tests/unit/stores/utilities/pool.spec.js
BenTopping: Yes there is an e2e test for it

seenanair (approved):

PR #1891, opened 2024-08-16
Y24-345 - Ont Pools store migration from axios to native fetch to reduce dependencies and increase maintainability.

stevieing:
Closes #

- Add extractIncludes and find method to Json API to allow finding of particular records for factory.
- Move ONT Pools store to v2 API
- remove ONT pools unit test data
- create OntAutotagFactory
- create OntLibraryFactory
- create OntPoolFactory
- create OntPlateFactory
- createOntTubeFactory
- move ONT Pool e2e tests to factories
- remove ONT Pool fixtures


src/api/JsonApi.js
```diff
@@ -305,6 +305,76 @@ const splitDataByParent = ({
...
+ * If the relationship is an array, then extract all the includes
+ * If the relationship is an object, then extract the include
+ * If the includes have their own relationships, then extract those
+ * @param {Object} relationships - the list of relationships to be extracted
+ * @param {Array} included - the list of included resources
+ * @param {Number} depth - the depth of the extraction
+ * @returns {Array} - the list of extracted includes
+ */
+const extractIncludes = (relationships, included, depth = 1) => {
+  // if we don't set a depth we get a stack overflow
```
BenTopping, line 319: This seems like a bit of a code smell, stack overflow typically implies there is an infinite recursion somewhere. I think it might be caused by false positives for the relationships check on line 342, not sure.

src/store/traction/ont/pools/actions.js
```diff
@@ -367,15 +368,16 @@ export default {
    */
   // For the component, the included relationships are not required
   // However, the functionality does not appear to work without them
+  // no unit tests?
```
BenTopping, line 371: Worth adding unit tests here?

BenTopping (changes requested):
Looks good. Pulled down locally and works fine.
I think some of the factories are quite large but I'm not sure if its worth spending time refactoring them further.

src/store/traction/ont/pools/actions.js
stevieing: It is covered bye e2e tests. We can add some at some point but I needed to stop somewhere.

src/api/JsonApi.js
stevieing: There is a specific reason which I will try and explain. Sometimes and includes will have relationships and we also need these.

The problem is you can sometimes get a circular relationship e.g. with tags and tag sets.

So a library will have a tag which will have a tag set which will have tags and so on.

There is no other way to distinguish other than having a depth. I have set it to 3. I couldn't think of any other way. Before I wasn' getting all of the data.

BenTopping: Ah I see, thanks for the explanation. Sounds like a tricky problem... As your comment suggests depth should be fine then 👍

stevieing:
> Looks good. Pulled down locally and works fine. I think some of the factories are quite large but I'm not sure if its worth spending time refactoring them further.

The data in the factories was copied directly from the json files so no bigger than previously. I did manage to remove over 4000 lines of code in this pull request.

The next stage would be to dynamically create the data but that is quite tricky so best left for another time.

BenTopping (dismissed):

src/api/JsonApi.js
seenanair: @stevieing My only comment was about number 3, which sounded like a magic number. Your explanation sounds good. Would it be good to that as a comment above so it won't create confusion in future?

seenanair (approved):

PR #1895, opened 2024-08-19
Develop into Master

stevieing:
Closes #

- Y24-228 - As a user (James W) I would like to add new label printing workflows 
- Y24-249 - [BUG] Filtering results for PacBio runs freezes page
- Various dependencies


src/views/pacbio/PacbioRunIndex.vue
```diff
@@ -167,12 +167,12 @@ const fields = ref([
...
   - `text`: The text to be displayed for the filter option.
 */
-const filterOptions = ref([
+const filterOptions = [
   { value: '', text: '' },
   { value: 'name', text: 'Name' },
   { value: 'state', text: 'State' },
   // Need to specify filters in json api resources if we want more filters
-])
+]
```
seenanair, line 175: I hope this doesn't have any implications on removing ref.

stevieing: I am not sure why but it looks like the ref was causing the problem @SHIV5T3R can you explain why?

BenTopping (approved):

src/views/pacbio/PacbioRunIndex.vue
BenTopping: the ref was breaking the filtering as the filter was expecting an array and receiving a ref object instead.

SHIV5T3R: My initial solution was
```js
filterOptions.filter(({ value }) => value == filter_value.value)[0]?.wildcard &&
```

But `filterOptions` doesn't really need to be a reference, it shouldn't have any implications

seenanair: Ok, thanks for the explanation. 👍

PR #1908, opened 2024-08-29
Y24-281: Filter not to include basecalls

### LINKED ISSUE #1898: Y24-281 - As long read ONT we would like to remove the 'basecalls' data type from the drop-down in the general reception as we only choose 'basecalls and raw data'
**User story**
As long read ONT we would like to remove the 'basecalls' data type from the drop-down in the general reception as we only choose 'basecalls and raw data'

Mary / Steve

Mary

**Acceptance criteria**
- [x] When I go to the general reception and select the ONT pipeline I should only see 'basecalls and raw data'
- [x] We do not need to remove this from the service as there will still be records associated.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

dasunpubudumal:
Closes #1898 

- Remove `basecalls` from ONT data types

### DIFF
--- a/src/components/shared/DataTypeSelect.vue
+++ b/src/components/shared/DataTypeSelect.vue
@@ -1,7 +1,7 @@
 <template>
   <traction-select
     id="data-type"
-    :model-value="value"
+    :model-value="dataType"
     data-attribute="data-type-list"
     :options="dataTypes"
     @update:model-value="handleInput"
@@ -12,18 +12,16 @@
 import useSWRV from 'swrv'
 import { filterByAttribute, mapAttribute } from '@/api/JsonApi'
 
+const encode = (value) => (value === undefined ? 'basecalls and raw data' : value)
 export default {
   name: 'DataTypeSelect',
   props: {
     pipeline: {
-      // Filter to only list data_types for a given pipeline, leave null
-      // to apply no filters
       type: String,
       required: false,
       default: null,
     },
-    value: {
-      // The data type, we use value to allow us to simply bind it with v-model
+    modelValue: {
       type: String,
       default: undefined,
     },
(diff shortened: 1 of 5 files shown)


src/components/shared/DataTypeSelect.vue
```diff
@@ -42,7 +42,9 @@ export default {
       return filterByAttribute(this.remoteDataTypes?.data || [], this.filters)
     },
     dataTypes() {
-      return mapAttribute(this.filteredDataTypes, 'name')
+      return mapAttribute(this.filteredDataTypes, 'name').filter(
```
BenTopping, line 45: We are already doing some filtering above so think it should belong in there. We can change filters on line 40 to 

`filters: ({ pipeline }) => (pipeline ? { pipeline, name: 'basecalls and raw data' } : {}),`

BenTopping (commented):
Just a small suggestion about the filtering. I think it would also be good to set the default value for the select box to 'basecalls and raw data' now that we only have one value.

BenTopping (commented):
Looks good. Can we also make basecalls and raw data the default value? Saves the users having to click and select since there is now only one option in the dropdown.

src/lib/receptions/index.js
```diff
@@ -12,7 +12,7 @@ const defaultRequestOptions = () => ({
   estimate_of_gb_required: null,
   number_of_smrt_cells: null,
   number_of_flowcells: null,
-  data_type: null,
+  data_type: 'basecalls and raw data',
```
BenTopping, line 15: I think this is probably fine but it means we will always send data_type, even when we don't want to, because we just [add all request data when building the payload](https://github.com/sanger/traction-ui/blob/develop/src/lib/receptions/sequencescapeUtils.js#L180). This will get filtered out in the service so it probably doesn't matter if its a lot of effort to change.

src/components/shared/DataTypeSelect.vue
```diff
@@ -1,56 +1,83 @@
...
     return { remoteDataTypes }
   },
+  data() {
+    return {
+      dataType: encode(this.modelValue),
+    }
+  },
   computed: {
+    // Same approach as src/components/shared/LibraryTypeSelect.vue
     filters: ({ pipeline }) => (pipeline ? { pipeline } : {}),
```
BenTopping, line 48: I think you need to add the filter back into here for 'basecalls and raw data'

dasunpubudumal: oh no..

BenTopping (approved):

PR #1918, opened 2024-09-05
Y24-160 ont runs axios to fetch

seenanair:
Closes # https://github.com/sanger/traction-ui/issues/1919

- Create ontRunsFactory
- Create ontInstrumentsFactory
- Change to v2 (fetch) in ontRoot, and ontRuns
- Refactor ONTRunIndex to use ontRoot Pinia store instead of vueX
- Refactor ONTRuns.spec to use the ontRoot Pinia store, v2 api and test factories
- Refactor all associated unit and e2e tests to use v2 and test factories
- Remove e2e fixture data files and unit data files associated with ontRuns and ontInstruments


src/views/ont/ONTRunIndex.vue
```diff
@@ -41,10 +41,8 @@
...
 import DownloadIcon from '@/icons/DownloadIcon.vue'
 import FilterCard from '@/components/FilterCard.vue'
-import useQueryParams from '@/composables/useQueryParams'
-
-import { createNamespacedHelpers } from 'vuex'
-const { mapActions, mapGetters } = createNamespacedHelpers('traction/ont')
+import useQueryParams from '@/composables/useQueryParams.js'
+import useOntRootStore from '@/stores/ontRoot.js'
 
 export default {
```
dasunpubudumal, line 47: This file looks different compared to `src/views/pacbio/PacbioPoolIndex.vue`. I believe both are using Pinia. Is it because `src/views/pacbio/PacbioPoolIndex.vue` is using new Vue features or is that the other way around?

seenanair: Both components are now using the Pinia store, but previously, OnRunIndex was using the VueX store for the ontRoot store. The difference between the two components is that PacbioPoolIndex uses a different syntax for Vue 3. PacbioPoolIndex was recently migrated from Vue 2 to Vue 3, so it uses the more modern <script setup>, which is a compiler macro that allows for less boilerplate code and includes improvements like automatic imports, among others.

tests/e2e/specs/ont/ont_run_view.cy.js
```diff
@@ -1,15 +1,27 @@
...
+
 describe('ONT Run page', () => {
   beforeEach(() => {
-    cy.intercept('/v1/ont/instruments', {
-      fixture: 'tractionOntInstruments.json',
+    cy.wrap(OntRunsFactory()).as('ontRunsFactory')
+    cy.wrap(OntInstrumentsFactory()).as('ontInstrumentsFactory')
+
+    cy.get('@ontInstrumentsFactory').then((ontInstrumentsFactory) => {
+      cy.intercept('GET', '/v1/ont/instruments', {
```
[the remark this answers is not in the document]
seenanair: Oh, ok good point. As we are specifying the 'GET' method, it should be fine I hope.

src/views/ont/ONTRunIndex.vue
dasunpubudumal: Interesting. Looks like it injects a bunch of code. Anyway thanks!

tests/e2e/specs/ont/ont_run_view.cy.js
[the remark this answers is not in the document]
stevieing: I have done this a few times and have been adding the method. When I have been working on this I have been running the Cypress tests locally with developer tools open to check if there are requests that are failing. e.g. in some cases I saw that printers were being fetched but not intercepted so were failing. I think it is best to ensure that all requests are handled even if the tests are passing.

tests/unit/components/ont/runs/ONTRunInstrumentFlowcells.spec.js
```diff
@@ -69,7 +70,7 @@ describe('ONTRunInstrumentFlowcells', () => {
     })
 
     it('must have instruments', () => {
-      const expected = mockInstruments.map((i) => {
+      const expected = Object.values(mockInstruments).map((i) => {
```
stevieing, line 73: Would this be better in the factory?

tests/unit/store/traction/ont/getters.spec.js
```diff
@@ -1,15 +1,18 @@
...
+const ontInstrumentsFactory = OntInstrumentsFactory()
+const ontRunsFactory = OntRunsFactory()
 
 describe('getters.js', () => {
   let state, instruments, runs
 
   beforeEach(() => {
-    instruments = new Response(Data.OntInstruments).deserialize.instruments
-    runs = new Response(Data.OntRuns).deserialize.runs
+    instruments = ontInstrumentsFactory.storeData.instruments
```
stevieing, line 14: Can you do `{ instruments, runs } = ontInstrumentsFactory.storeData`

stevieing (changes requested):
Nice. Just a couple of suggestions. Might also be worth running jsdoc to see the docs. I think there might be a couple of things that might not work correctly. Some of your methods do not have signatures.

tests/factories/OntInstrumentsFactory.js
```diff
@@ -0,0 +1,50 @@
+import BaseFactory from './BaseFactory.js'
+import { dataToObjectById } from '../../src/api/JsonApi.js'
+/*
```
stevieing, line 3: I think having the single * means that jsdoc doesn't pick it up.

tests/unit/components/ont/runs/ONTRunInstrumentFlowcells.spec.js
seenanair: Yes, it is used in a couple of places. But we need this as object as well in other places, so an additional method to return as array in factory might be good I hope.

tests/unit/store/traction/ont/getters.spec.js
seenanair: We may not be able to destructure it this way it as they are from different factories, instruments from ontInstrumentsFactory.storeData and runs from ontRunsFactory.storeData

stevieing: Ah. Didn't notice that.

stevieing (approved):

PR #1933, opened 2024-09-12
Y24-263 - replacing the static json workflow file by API response

sabrine33:


src/stores/printing.js
```diff
@@ -90,5 +100,20 @@ export const usePrintingStore = defineStore('printing', {
...
       return { success, data, errors }
     },
+
+    /**
+     * fetches the list of workflows from traction
+     * @returns {Object}  {success: Boolean, data: Object, errors: Array} Is the fetch successful? The data and any errors
+     */
+    async fetchWorkflows() {
+      const rootStore = useRootStore()
+      const request = rootStore.api.v1.traction.workflows
```
seenanair, line 110: Also, we are currently in the middle of migrating from Axios to Fetch, where api.v1 uses Axios and api.v2 uses Fetch. Therefore, it might be better to use api.v2 for this, since it is a new addition.

src/stores/printing.js
```diff
@@ -90,5 +100,20 @@ export const usePrintingStore = defineStore('printing', {
       }
       return { success, data, errors }
     },
+
+    /**
+     * fetches the list of workflows from traction
+     * @returns {Object}  {success: Boolean, data: Object, errors: Array} Is the fetch successful? The data and any errors
+     */
+    async fetchWorkflows() {
```
seenanair, line 108: Is  printing store a good place to keep it?Maybe the method can be placed in helper file as well for e.g lib/LabelPrintinHelpers.js with workflows stored inside the component as it is only used in one place.

seenanair (commented):
Good start on UI 👍 
Just shared a few thoughts.

src/components/labelPrinting/LabelPrintingForm.vue
```diff
@@ -250,15 +255,19 @@ const onReset = () => {
...
-// if no printers are in the store, fetch them
+// if no printers or workflows in the store, fetch them
 // if there are printers in the store, return success prevents error in DataFetcher
 // @returns {Promise} - Promise
-const fetchPrinters = async () => {
-  if (printingStore.printers().length === 0) {
-    return await printingStore.fetchPrinters()
-  } else {
-    return { success: true }
+const fetchResources = async () => {
```
stevieing, line 261: I'm not sure this is the ideal way to deal with thi. @BenTopping can you have a look. From what I remember we should use the DataFetcher. Doing it this way could cause issues. Also we are always returning true whatever.

src/stores/printing.js
```diff
@@ -90,5 +98,20 @@ export const usePrintingStore = defineStore('printing', {
...
```
stevieing, line 106: This probably needs a bit of work. The standard is to return the workflows with the steps and store them in 2 separate pieces of state. You can then bring them together in the component. I can go through this with you.

src/api/JsonApi.js
```diff
@@ -305,6 +305,49 @@ const splitDataByParent = ({
   }, {})
 }
 
+/**
```
stevieing, line 308: I am not sure this is necessary. It is better to use side loading for the workflows which makes it more maintainable. I can go through it with you.

stevieing (changes requested):
Good attempt. It is better if when we are retrieving data from the store or service that we follow the same pattern as other resourcesas it is more efficient and maintainable. Happy to talk through it with you.

stevieing (approved):
Good work.

PR #1978, opened 2024-09-27
Y24-143 - LabWhere API support

### LINKED ISSUE #1774: Y24-143 - As developers we need to create a software library for managing locations from labwhere that we can use in traction so that long read user can see where their samples are without having to go to labwhere
**User story**
As developers we need to create a library locations from labwhere so that we can use in traction so that long read user can see where their samples are without having to go to labwhere

Steve / Danni / James

Steve / Danni / James

**Acceptance criteria**
- [ ] A new method of data retrieval as labwhere does not use JSON API
- [ ] A separate library which can be reused in multiple views
- [ ] Fetch locations from labwhere for multiple barcodes
- [ ] Retrieval of the co-ordinate for a barcode as they are working in boxes

**Dependencies**
- #714 

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1774 

- Adds LabWhere service directory.
- Adds initial helper methods for retrieving LabWhere locations given a labware barcode. 
- Adds FetchWrapper, a basic class to handle non JSON-API compliant fetches in a standard way.


seenanair (approved):
Looks good. 
As we discussed, FetchWrapper can be further extended to support additional use cases, such as handling other content types, as part of the corresponding stories.

src/api/FetchWrapper.js
```diff
@@ -0,0 +1,55 @@
+/**
```
stevieing, line 1: Advisory: It might be worth rethinking using a class here.

I have tried using them before and they can quickly get quite difficult to manage and they do not behave like traditional OO classes.

BenTopping: Replaced with a function object in https://github.com/sanger/traction-ui/pull/1978/commits/2af437eec165b1484207ad35073aac11852a480d.

stevieing (approved):
Nice. Looking good.

PR #1983, opened 2024-09-30
Y24-353 Fix for Missing data on edit for ONT runs

seenanair:
Closes #

Update the 'include' parameter in the 'find' and 'create' method of createRequest so that destructuring of parameters works as in v1

Correct unit test cases to replicate the way it is called from stores

### DIFF
--- a/tests/e2e/specs/ont/ont_runs_view.cy.js
+++ b/tests/e2e/specs/ont/ont_runs_view.cy.js
@@ -42,10 +42,7 @@ describe('ONT Runs view', () => {
         body: ontRunFactory.content,
       })
     })
-    cy.intercept('v1/ont/pools?filter[barcode]=TRAC-1-2', {
-      statusCode: 200,
-      fixture: 'tractionOntPool.json',
-    })
+  
     cy.intercept('/v1/ont/pools?include=tube,libraries.tag,libraries.request', {
       fixture: 'tractionOntPools.json',
     })
--- a/tests/e2e/specs/ont/ont_runs_view.cy.js
+++ b/tests/e2e/specs/ont/ont_runs_view.cy.js
@@ -42,7 +42,7 @@ describe('ONT Runs view', () => {
         body: ontRunFactory.content,
       })
     })
-  
+
     cy.intercept('/v1/ont/pools?include=tube,libraries.tag,libraries.request', {
       fixture: 'tractionOntPools.json',
     })
--- a/tests/e2e/specs/ont/ont_runs_view.cy.js
+++ b/tests/e2e/specs/ont/ont_runs_view.cy.js
@@ -1,4 +1,5 @@
 import OntRunsFactory from '../../../factories/OntRunsFactory.js'
+import OntRunFactory from '../../../factories/OntRunFactory.js'
 
 describe('ONT Runs view', () => {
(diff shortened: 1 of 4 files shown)


stevieing (changes requested):
Do we need an addition to the e2e test for creating and editing an ONT run to ensure the data is shown? This will guard against a similar issue in the future.

seenanair:
> Do we need an addition to the e2e test for creating and editing an ONT run to ensure the data is shown? This will guard against a similar issue in the future.

I actually tried it, but it didn't work as I expected. Might need bit more time looking into it.

stevieing:
> > Do we need an addition to the e2e test for creating and editing an ONT run to ensure the data is shown? This will guard against a similar issue in the future.
> 
> I actually tried it, but it didn't work as I expected. Might need bit more time looking into it.

OK. Let's timebox it and if needed roll it into it's own story. It may well be an issue with the page loading. The ONT run page is very slow to load. We do have a technical debt story somewhere.

seenanair:
> > > Do we need an addition to the e2e test for creating and editing an ONT run to ensure the data is shown? This will guard against a similar issue in the future.
> > 
> > 
> > I actually tried it, but it didn't work as I expected. Might need bit more time looking into it.
> 
> OK. Let's timebox it and if needed roll it into it's own story. It may well be an issue with the page loading. The ONT run page is very slow to load. We do have a technical debt story somewhere.

Ok. Also, it would be good to add similar e2e test cases for Pacbio Pool, Runs etc. for the same reason

stevieing:
> > > > Do we need an addition to the e2e test for creating and editing an ONT run to ensure the data is shown? This will guard against a similar issue in the future.
> > > 
> > > 
> > > I actually tried it, but it didn't work as I expected. Might need bit more time looking into it.
> > 
> > 
> > OK. Let's timebox it and if needed roll it into it's own story. It may well be an issue with the page loading. The ONT run page is very slow to load. We do have a technical debt story somewhere.
> 
> Ok. Also, it would be good to add similar e2e test cases for Pacbio Pool, Runs etc. for the same reason

I have created a story  #1984

stevieing (dismissed):

tests/e2e/specs/ont/ont_runs_view.cy.js
```diff
@@ -23,4 +24,35 @@ describe('ONT Runs view', () => {
...
+    cy.intercept('/v1/ont/pools?include=tube,libraries.tag,libraries.request', {
+      fixture: 'tractionOntPools.json',
+    })
+
+    cy.visit('#/ont/runs')
+    cy.get('#editRun-2').click()
+
+    // Check that the URL is correct
+    cy.url().should('include', '#/ont/run/2')
+    cy.get('#flowcell-id-1').invoke('val').should('eq', 'ABC1234')
```
stevieing, line 55: Advisory. We should probably change this so it relates to the store data but I suggest we do that at another time and get this story deployed.

stevieing (approved):

tests/e2e/specs/ont/ont_runs_view.cy.js
seenanair: Ok 👍

Issue #1773, opened 2024-06-20
Y24-141 - As a team lead (Danni) I would like to have a reception so that I can scan tubes and plates into locations without having to visit labwhere and to speed up the process


stevieing:
**User story**
As a team lead (Danni) I would like to have reception so that I can scan tubes and plates into locations without having to visit labwhere and to speed up the process

**Who are the primary contacts for this story**
Steve / Danni / James

**Who is the nominated tester for UAT**
Steve / Danni / James

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] A location reception page in traction
- [x] An input for the User swipe card id / barcode
- [x] An input for the location barcode
- [x] An input for the start position
- [x] An input for the barcodes
- [x] A counter for how many barcodes are to be scanned
- [x] A big button to submit the scan
- [x] Feedback if the scan fails
- [x] Feedback if the scan succeeds
- [x] The locations to be updated in labwhere
- [x] Validation for duplicate barcodes

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #714

**Additional context**

PR #1986, opened 2024-10-01
Y24 141 labwhere reception to scan tubes and plates into locations

seenanair:
Closes https://github.com/sanger/traction-ui/issues/1773

- Adds Labwhere reception page to store labwhere barcodes 
   - This includes a form interface for users to scan in/out labware barcodes to/from a specified location in Labwhere.
   -  The form includes fields for: User barcode or swipecard, Location barcode, Start position (optional), Labware barcodes
   - The component validates the input fields for 'User barcode', and 'Labware barcodes'
       and displays error messages for any invalid or missing data when the form is submitted or when the input fields are 
        updated.
   - The 'Start position' field is optional and does not require validation.
   -  Upon successful validation, the form data is submitted to the Labwhere API to store the barcodes in the specified 
         location.
   - It displays a success message if the barcodes are stored successfully, or an error message if the submission fails.

- Adds `scanBarcodesIntoLabwhereLocation` helper method for storing barcodes to Labwhere location using Labwhere `scans` api
- Updates `FetchWrapper` class to have content_type parameter

Note : @BenTopping @stevieing 
During testing, it was realised that there is currently no validation for barcodes in LabWhere, allowing any text entered as a labware barcode to be stored in a location.  This raises the question of whether we need separate validation in Traction for invalid barcodes using the Sequencescape API, which requires further discussion and review.
(In similar cases, such as when using `scans` api in AssetAudits, barcode validation is handled through the Sequencescape API.)


tests/unit/views/LabWhereReception.spec.js
```diff
@@ -0,0 +1,139 @@
...
+
+    const barcodeInput = wrapper.find('#labware_barcodes')
+    await barcodeInput.setValue('barcode1')
+
+    await wrapper.vm.$nextTick()
+    expect(wrapper.vm.errors).toEqual({
+      user_code: 'User code is required',
+    })
+
+    const [userInput] = wrapper.findAll('input')
```
BenTopping, line 75: Worth using a test-id or data attribute to make input more discoverable?

src/router.js
```diff
@@ -53,6 +54,12 @@ const router = createRouter({
       meta: { pipeline: 'Reception' },
       component: GeneralReception,
     },
+    {
+      path: '/labwhere-reception',
+      name: 'LabwhereReception',
+      meta: { page: 'Labwhere Reception - Scan In/Out' },
```
BenTopping, line 60: LabWhere*

BenTopping (changes requested):
Good start, tested locally and the page looks good and works well with nice error handling 👌 

Couple things:
- application/json should work as the content-type for the scans api. Ive tested locally with a basic CURL and seems to work so might be a setup issue for the CORS error.
- Im not sure the preview area of the page adds much. Maybe a text highlight in red for duplicate barcodes and an error preventing submission if there are duplicates would be better.
- A clear button to remove the labware location, position and barcodes might be useful if they are doing a lot of labware at once (not sure if this will happen but worth thinking about).

seenanair:
> Good start, tested locally and the page looks good and works well with nice error handling 👌
> 
> Couple things:
> 
> * application/json should work as the content-type for the scans api. Ive tested locally with a basic CURL and seems to work so might be a setup issue for the CORS error.

Oh, interesting. I tried it many different ways, but never worked for me with CORS error.

> * Im not sure the preview area of the page adds much. Maybe a text highlight in red for duplicate barcodes and an error preventing submission if there are duplicates would be better.

Following was my thought to do a preview area. 'Labware count' is specified as a requirement by the stakeholder, and I believe it would be useful to display it as preview text. Additionally, since we are not validating against an empty location barcode, which could potentially cause a reverse effect (scanning out) on normal operations, it is important to display the intended action's purpose somewhere for confirmation.

> * A clear button to remove the labware location, position and barcodes might be useful if they are doing a lot of labware at once (not sure if this will happen but worth thinking about).
A 'Reset' button was initially there but in attempt to make it similar to LabWhere scan in page, I removed it.

seenanair:
> > Good start, tested locally and the page looks good and works well with nice error handling 👌
> > Couple things:
> > 
> > * application/json should work as the content-type for the scans api. Ive tested locally with a basic CURL and seems to work so might be a setup issue for the CORS error.
> 
> @BenTopping Oh, interesting. I tried it many different ways, but never worked for me with CORS error.
> 
> > * Im not sure the preview area of the page adds much. Maybe a text highlight in red for duplicate barcodes and an error preventing submission if there are duplicates would be better.
> 
> Following was my thought to do a preview area. 'Labware count' is specified as a requirement by the stakeholder, and I believe it would be useful to display it as preview text. Additionally, since we are not validating against an empty location barcode, which could potentially cause a reverse effect (scanning out) on normal operation of scanning in, thought it is important to display the intended action's purpose somewhere for confirmation.
> 
> > * A clear button to remove the labware location, position and barcodes might be useful if they are doing a lot of labware at once (not sure if this will happen but worth thinking about).
> 
> A 'Reset' button was initially there but in attempt to make it similar to LabWhere scan in page, I removed it. :)

BenTopping (approved):

PR #2003, opened 2024-10-17
Y24-344 - QC Results Upload migration to fetch.

### LINKED ISSUE #1966: Y24-344 - QcResultUpload service migration from axios to native fetch to reduce dependencies and increase maintainability.
**Describe the Housekeeping**

The goal of this task is to improve the codebase by reducing dependencies and increasing maintainability. Specifically, we will refactor the QcResultUpload service to use fetch instead of axios, and create test factories to replace fixture data.

1. **Refactor the QcResultUpload service so that it uses fetch instead of axios** : The primary reason for switching from axios to fetch is to minimize the use of external libraries wherever possible. By using the native fetch API, we reduce dependency on third-party libraries, which simplifies the codebase and makes it easier to maintain in the long run.
  
2. **Create and use test factories instead of test files** : Introducing test factories instead of manually maintaining test files and fixtures helps avoid duplication of fixture data between unit and e2e tests. This makes the testing process more consistent, maintainable, and easier to update when changes are made in the codebase. The factory will generate standardized test data across different tests, streamlining the testing process and reducing potential errors caused by inconsistent data.

The major tasks involve:

- [ ] Create a factory (example: https://github.com/sanger/traction-ui/pull/1756/files#diff-c5d273e6d7069f315f8d62842f5753d1599be7c0799f26fb6199392203007f2c)
- [ ] plug factory into tests ( doing this first will ensure that when we change the code we kow what the failures are related to)
- [ ] rationalise any use of test data e.g. reduce places where it is used. Standardise where possible
- [ ] Change handleResponse from v1 to v2 in store or library
- [ ] some stores use the old response type with deserialize - so you would need to move over to the new response type and use side loading
- [ ] make sure you only use relative namespacing for imports because Cypress does not have access to @.
- [ ] make sure you use file extensions e.g. .js at the end of the import name as this improves performance.
- [ ] Change request from v1 to v2
- [ ] Fix unit tests
- [ ] Run tests for any components where store or library is used and make any changes and fix
- [ ] Run tests for any views where store or library is used and make any changes and fix
- [ ] Fixe e2e tests
- [ ] Remove use of fixtures where store or library is used and use factory
- [ ] Remove data file and fixture

**Blocking issues**
Describe any other issues or tickets that may be blocking this change.

**Additional context**
[Y24-138 - As developers we want to move over all requests from Axios to fetch to remove a dependency and make the codebase easier to maintain](https://github.com/sanger/traction-ui/issues/1767)

BenTopping:
Closes #1966 

- Changes QcResultsUploadForm to use v2 api
- Updates QcResultsUpload service to use fetch response objects and handle errors more cleanly.


tests/factories/QcResultUploadFactory.js
```diff
@@ -0,0 +1,27 @@
+import fs from 'fs'
+import BaseFactory from './BaseFactory.js'
+
+const defaultCsv = fs.readFileSync('./tests/data/csv/qc-results-upload.csv', 'utf8')
```
stevieing, line 4: Could we just put the file in here rather than having to load it synchronously?

That way we can get rid of the data file and use it in the e2e tests.

At the moment I do not think this is usable in the e2e tests as readFileSync is not available in the browser.

stevieing (changes requested):
Just some questions about the use of the file.

I think it would be better to put the csv file directly in the Factory. That way we can get rid of the data file and the e2e fixtures.

tests/factories/QcResultUploadFactory.js
stevieing: Ok. I can see what you are saying.

I plan to look at this again as there are places we are using csvs when I think we could create them on the fly by using a temp file or something similar. e.g. pool creation.

stevieing (approved):

PR #2011, opened 2024-10-22
Y24-338 - Pacbio libraries store migration from axios to native fetch

stevieing:
Closes #

- 


src/components/pacbio/PacbioLibraryCreate.vue
```diff
@@ -122,11 +122,13 @@ const toggleDisplayCreatePanel = () => {
 
 /**
  * @method createLibrary
- * @description Creates a new library by calling the createLibraryInTraction method from the 'pacbioLibraries' store.
+ * @description Creates a new library by calling the createLibrary method from the 'pacbioLibraries' store.
  * @returns {void} Displays a success message if the library is created successfully, otherwise displays a failure message.
  */
 const createLibrary = async () => {
-  const { success, barcode, errors } = await librariesStore.createLibraryInTraction(
+  console.log('started')
```
BenTopping, line 129: remove console.log

src/stores/utilities/pacbioLibraries.js
```diff
@@ -0,0 +1,45 @@
...
+}) => {
+  const requiredAttributes = {
+    template_prep_kit_box_barcode,
+    tag_id,
+    concentration,
+    volume,
+    insert_size,
+  }
+
+  return id
```
BenTopping, line 18: Another way of writing this would be define the shared structure in a variable and then add a conditional after which adds the id-relevant data. Might prevent some duplication but not required.

src/stores/pacbioLibraries.js
```diff
@@ -194,28 +192,9 @@ export const usePacbioLibrariesStore = defineStore('pacbioLibraries', {
...
-              insert_size: libraryFields.insert_size,
-              tag_id: libraryFields.tag_id,
-            },
-          },
-        },
-      }
-      const promise = request.update(body)
+      const request = rootStore.api.v2.traction.pacbio.libraries
+      const payload = libraryPayload(libraryFields)
+      const promise = request.update(payload)
```
BenTopping, line 197: much cleaner 👍

tests/unit/components/pacbio/PacbioLibraryCreate.spec.js
```diff
@@ -75,21 +74,16 @@ describe('PacbioLibraryCreate.vue', () => {
...
+      cmp = wrapperObj.vm
     })
 
     it('will have a Create Library button component', () => {
       const button = wrapper.findComponent('#pacbioLibraryCreate')
-      expect(button.element).toBeTruthy()
       expect(button.element.disabled).toBe(false)
     })
 
-    it('will have an modal component', () => {
```
BenTopping, line 87: How was this modal check passing before 🤔 ?

tests/unit/components/pacbio/PacbioLibraryCreate.spec.js
```diff
@@ -99,50 +93,52 @@ describe('PacbioLibraryCreate.vue', () => {
...
+      expect(cmp.createLibrary).toBeCalled()
     })
 
     it('should not display library form when the cancel button is clicked', async () => {
       wrapper.find('#pacbioLibraryCreate').trigger('click')
       await nextTick()
-      expect(modal.isDisplayLibraryForm).toBeTruthy()
+      expect(cmp.isDisplayLibraryForm).toBeTruthy()
       wrapper.find('#cancel-btn').trigger('click')
       await nextTick()
```
BenTopping, line 109: this test is a bit weird, it checks the displayform variable is true as a means of checking the library form is displayed, then checks the library form element does not exist as a means of checking the library is not displayed. I think both checks should either be for the '#libraryForm' element or for the modal variable (or both).

tests/unit/components/pacbio/PacbioLibraryCreate.spec.js
```diff
@@ -99,50 +93,52 @@ describe('PacbioLibraryCreate.vue', () => {
...
       expect(wrapper.element.querySelector('#libraryForm')).toBeNull()
     })
 
+    // skipping to check if the ci runs and e2e tests pass
```
BenTopping, line 113: can remove comment.

BenTopping (changes requested):

tests/unit/components/pacbio/PacbioLibraryCreate.spec.js
stevieing: It was stubbing modal.createLibraryInTraction. Because that is a method on the store which is available on the modal it was working. When I changed it to be the same name it was then stubbing the modal method rather than the store method.

My suspicion is that is why the method name was changed.

src/stores/utilities/pacbioLibraries.js
stevieing: I did think about that but wasn't sure. If it was just the id that would be easy but I got a bit stuck with how to add the pacbio_request_id into the attributes. If you have a solution let me know. I am probably missing the obvious.

tests/unit/components/pacbio/PacbioLibraryCreate.spec.js
stevieing: It is weird. I almost feel like removing all 3 of these tests? as they are testing implementation and probably sufficiently tested elsewhere and in the e2e tests.

BenTopping: Happy to have them removed if they are covered in e2e 👍

tests/unit/stores/pacbioLibraries.spec.js
```diff
@@ -12,80 +34,49 @@ describe('usePacbioLibrariesStore', () => {
     setActivePinia(pinia)
   })
 
-  describe('state', () => {
-    let libraries
-    beforeEach(() => {
-      libraries = new Response(Data.TractionPacbioLibrary).deserialize.libraries
+  describe('#libraryPayload', () => {
```
BenTopping, line 37: Wondering if this should be moved to a separate utilities test.

stevieing: I did think about that but because it was only one function not sure it is worth it. If the utilities file expands then definitely. I wanted to put this function in the store but it doesn't make it easy to test.

BenTopping (approved):

PR #2021, opened 2024-10-25
Fix PacBio pool creation csv upload failure 

sabrine33:


tests/e2e/specs/pacbio/pacbio_pool_create.cy.js
sabrine33, line 1: I updated the test to mimic the actual file upload use case, ensuring that the error is captured and future regressions are prevented.

stevieing (approved):

PR #2044, opened 2024-11-07
1962-Y24-340 - pacbio pool create store migration from axios to native fetch to reduce dependencies and increase maintainability

stevieing:
Closes #

- 


tests/unit/stores/utilities/pool.spec.js
```diff
@@ -296,4 +272,377 @@ describe('pool', () => {
...
+            data: [
+              {
+                type: 'aliquots',
+                id: '315',
+              },
+            ],
+          },
+        },
+      },
+    ]
```
seenanair, line 632: Same question as above about data

tests/unit/stores/utilities/pool.spec.js
```diff
@@ -296,4 +272,377 @@ describe('pool', () => {
...
+          },
+          pool: {
+            data: {
+              type: 'pools',
+              id: '6011',
+            },
+          },
+        },
+      },
+    ]
```
seenanair, line 503: Do we need to keep this data over here or can it be moved to factory?

src/stores/pacbioPoolCreate.js
```diff
@@ -594,18 +600,20 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
      * // Populate used_aliquots from a pool
      * const result = await populateUsedAliquotsFromPool(1);
      * console.log(result); // { success: true, errors: [] }
+     * This method actually fetches the pool so should be renamed. It was difficult to work out which method created the pool,
+     * fetchPool
      */
```
seenanair, line 605: Good point. Do you think it is good to rename as part of this PR?

src/stores/pacbioPoolCreate.js
```diff
@@ -618,35 +626,32 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
+        this.resources.tubes = assignLibraryRequestsToTubes({
+          libraries: this.resources.libraries,
+          requests: this.resources.requests,
+          tubes,
         })
+
+        // Get the pool tube and remove it from tubes list.
+        // There must be a reason why it is done this way?
+        this.tube = this.resources.tubes[data.relationships.tube.data.id]
+        delete this.resources.tubes[data.relationships.tube.data.id]
```
seenanair, line 653: I might be wrong, but is it that we are removing the pool tube so that it does not appear in the library tube list in places it is used, maybe  the pool edit page?

seenanair (commented):
Looks good. Just a few questions.

tests/unit/stores/utilities/pool.spec.js
stevieing: I think it needs to stay here because the factory uses those methods.

I have taken care to cut down the data as much as possible.

It also acts as a test. If we make any changes then the factory will break.

src/stores/pacbioPoolCreate.js
stevieing: I think the best thing to do would be to refactor it and move the post fetch processing elsewhere.

Probably not something I want to do as part of this.

We can talk about it when we do some aliquot refactoring.

seenanair (approved):

src/stores/pacbioPoolCreate.js
stevieing: That was my suspicion.

In which case I think the best course of action would be to rejig the store so that we can add separation for the different types of tubes rather than adding them to the resources and then removing them.

seenanair: Yes, agree. I don't think the store data should be tweaked for that purpose.

tests/unit/stores/utilities/pool.spec.js
sabrine33: @stevieing can we get the data from the factories instead of using fixture here ? we are not testing the factories here, so I think it should be fine, or may be I am missing something ?

stevieing: I did think about that but the factories use these methods so it would not make sense. It was also really difficult to make sense of what was going on using the factory data.

I think that these kind of methods sit better in the aliquot code so we can look at it as part of our review. I have noticed it all seems to relate to aliquot technical debt which it would be good to reduce and maybe remove these methods altogether.

I have cut down the data as much as possible and it was much easier to test like this as these methods are quite confusing.

sabrine33 (approved):

PR #2060, opened 2024-11-15
add scan in functionality in the general reception page

sabrine33:


BenTopping (commented):
Nice work, couple small comments. Could you also add a cypress test for this.

src/components/reception/SwipeCard.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <fieldset>
+    <traction-heading level="4" show-border>User barcode or swipecard</traction-heading>
+    <traction-muted-text class="flex justify-left">{{ description }}</traction-muted-text>
+    <traction-field-error data-attribute="user-code-error" :error="error">
+      <traction-input
+        id="userCode"
```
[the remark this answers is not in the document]
sabrine33: Yes, I was about to mention this, actually. In other systems (Limber and SS), they are, but this is how it is currently implemented in Traction. I'm not sure if it is ok to make it sensitive without user consent?

BenTopping (approved):

tests/unit/components/reception/SharedTestsForImportAndScanIn.js
```diff
@@ -0,0 +1,166 @@
...
+
+      it('calls created reception functionality ', async () => {
+        expect(mockedCreateReception).toBeCalledWith(
+          tractionReceptionsCreate,
+          foundBarcodes,
+          attributes,
+        )
+      })
+
+      it('create response messages accordingly ', async () => {
```
stevieing, line 101: Having a test just to say something is called is probably not necessary. As long as the result of that call is being tested elsewhere.

tests/unit/components/reception/SharedTestsForImportAndScanIn.js
```diff
@@ -0,0 +1,166 @@
...
+
+      it('does not call scanIn LabWhere functionality', () => {
+        expect(scanBarcodesInLabwhereLocation).not.toHaveBeenCalled()
+      })
+    })
+  })
+
+  describe('Successful import', () => {
+    describe('Successfully import and scan in to LabWhere ', () => {
+      let mockedCreateReception
```
stevieing, line 60: This is more of a statement. You are pretty much mocking everything. Generally the only thing you need to mock are API calls as they are the only things you can't test directly. It would be worth double checking if everything needs to be mocked. It can cause false positives and things might not get tested correctly.

tests/unit/components/reception/SharedTestsForImportAndScanIn.js
```diff
@@ -0,0 +1,166 @@
...
+        wrapper.vm.reception.fetchFunction.mockRejectedValue('Failed fetch')
+        await wrapper.vm.fetchLabware()
+        wrapper.vm.reception.fetchFunction = vi.fn()
+        vi.spyOn(Reception, 'createReceptionResource').mockImplementation(() => {})
+      })
+      it('shows error alert message', () => {
+        expect(mockShowAlert).toHaveBeenCalledWith('Failed fetch', 'danger')
+      })
+
+      it('does not call create resource functionality', async () => {
```
stevieing, line 28: Again you are just testing something has not been called. You should really only test the output which looks like it is already being tested.

src/views/GeneralReception.vue
```diff
@@ -160,6 +190,31 @@ const pipeline = ref(reception.value.pipelines[0])
...
+
+const workflowOptions = computed(() => [
+  { value: '', text: '' }, // Empty option
+  ...Object.values(WorkflowsLocations).map((workflow) => ({
+    value: workflow.barcode,
+    text: workflow.name,
+  })),
+])
+
+const workflowLocation = computed(() => {
```
stevieing, line 205: what is this doing?

src/components/reception/MultiplexedLibraryBarcode.vue
```diff
@@ -135,6 +139,18 @@ const props = defineProps({
     type: Object,
     default: () => ({}),
   },
+  additionalDetails: {
```
stevieing, line 142: Looks like this is duplicated in several places? Not sure if it is worth abstracting out but worth considering.

stevieing (changes requested):
A few comments and questions.

I have been doing a big review of our tests so have noticed quite a few things elsewhere so worth checking here.

If something is covered in the e2e tests then it doesn't need to be covered in unit tests.

A unit test should test output so it is worth checking if it actually does what it is supposed to do rather than mocking everything out.

In unit tests just testing something is called is generally unneccessary. You should be testing what comes out.

src/components/reception/MultiBarcode.vue
```diff
@@ -119,6 +123,18 @@ const props = defineProps({
     type: Object,
     default: () => ({}),
   },
+  additionalDetails: {
```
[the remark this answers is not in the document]
sabrine33: to display additional details within the components - I can rename it to something more specific for this use case as " workflowLocation"

src/components/reception/MultiplexedLibraryBarcode.vue
sabrine33: MultiplexedLibraryBarcode and The MultiBarcode component have the same defined properties as they are injected into the main component dynamically

tests/unit/components/reception/SharedTestsForImportAndScanIn.js
```diff
@@ -0,0 +1,120 @@
...
+        wrapper.vm.reception.fetchFunction.mockRejectedValue('Failed fetch')
+        await wrapper.vm.fetchLabware()
+        wrapper.vm.reception.fetchFunction = vi.fn()
+        vi.spyOn(Reception, 'createReceptionResource').mockImplementation(() => {})
+      })
+      it('shows error alert message', () => {
+        expect(mockShowAlert).toHaveBeenCalledWith('Failed fetch', 'danger')
+      })
+
+      it('does call scan in LabWhere functionality', () => {
```
stevieing, line 29: I think you can take this test out. You have already established that it does not scan.

stevieing (approved):
All good. 1 minor comment.

stevieing (approved):

Issue #2007, opened 2024-10-21
Y24-412 - As developers we need to create state management so that Long Read (Michelle) can create multiple libraries at once


stevieing:
**User story**
As developers we need to create state management so that Long Read (Michelle) can create multiple libraries at once

**Who are the primary contacts for this story**
Steve / Ben / Seena

**Who is the nominated tester for UAT**
Michelle / PSD

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] The action (createLibraryBatch) should receive a csv delimited file in the format of the attached file
- [ ] The action should receive a tag set specified by the user
- [ ] The action should build a new library for each sample
- [ ] All of the samples should be retrieved from the service by sample name to ensure they exist
- [ ] If the samples do not exist then the action should return an error
- [ ] The tag set should be returned by the service
- [ ] The tags should be checked to see if they exist within the tag set and if they are unique
- [ ] Each library should be validated
- [ ] If the data is correct then a request should be sent to the service
- [ ] If there are errors then these should be returned from the action
- [ ] If the libraries are created successfully then they should be returned from the action with the tube barcodes
- [ ] There should be a PacbioBatchLibraryFactory to handle the data
- [ ] There should be a fetchBatchLibraries action to retrieve list of batches

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

[x24 traction library csv example.csv](https://github.com/user-attachments/files/17458691/x24.traction.library.csv.example.csv)

PR #2064, opened 2024-11-19
Y24-412 Create state management for multiple libraries

seenanair:
Closes https://github.com/sanger/traction-ui/issues/2007

- Introduction of a pacbioLibraryBatchesStore:  A new store to manage the creation of library batches.

    createLibraryBatch Method: 
          A method to create a library batch, involving the following actions:
          -         Accepts a CSV file and a tagset as props.
          -         Ensures the uniqueness of tags.
          -         Validates the CSV data, checking for missing fields, the existence of sources, and whether the tags exist within the provided tagset.
          -         Validation and Library Creation:
                         If the data is validated successfull: Creates libraries and returns the corresponding data.
                         If validation fails: Returns an error.

- Tagset Association:  All tags within a batch will belong to a single tagset, which will be provided as a prop.

- Introduction of PacbioBatchLibraryFactory: A factory to handle and process the data.


tests/unit/stores/ontRoot.spec.js
```diff
@@ -122,7 +122,7 @@ describe('useOntRootStore', () => {
         const rootStore = useRootStore()
         const get = vi.fn()
         get.mockRejectedValue(failedResponse)
-        rootStore.api.v1 = { traction: { ont: { instruments: { get } } } }
+        rootStore.api.v2 = { traction: { ont: { instruments: { get } } } }
```
BenTopping, line 125: has ont root store changed?

BenTopping (commented):
Good start. I think there might be some more complexity around tag names and tag validation. Happy to chat.

src/lib/csv/pacbio.js
```diff
@@ -71,8 +74,23 @@ const eachRecord = (csv, callback) => {
     trim: true,
     info: true,
     cast,
-    onRecord: filterUndefinedValues,
-  }).forEach(callback)
+  }).map(filterUndefinedValues)
+
+  let retRecords = []
```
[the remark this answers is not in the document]
seenanair: As we discussed yesterday, the eachRecord function has been refactored to ensure that existing cases are not affected. Currently, eachRecord does not return any values, nor is it used in a validation mode where a failure to return a value would halt further processing.
Therefore, if eachRecord does not return a value (as is the case in existing implementations), the method will continue executing and simply return the parsed CSV records. 
Additionally, I have not converted it into a reduce function because the current implementation allows for an early return in case of invalid results.
I hope this makes sense.

src/stores/pacbioLibraryBatches.js
```diff
@@ -0,0 +1,150 @@
...
+      if (!csvFile) {
+        return { success: false, errors: ['csvFile is required'] }
+      }
+      try {
+        // Fetch the tags and requests
+        const { requests, tags } = await fetchTagsAndRequests(tagSet)
+
+        // Read the CSV file and validate the records
+        const csv = await csvFile.text()
+        const error = hasDuplicateTags(csv)
```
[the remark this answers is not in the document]
seenanair: As Mitchelle confirmed, we will be using only unique tags.

tests/unit/stores/ontRoot.spec.js
seenanair: No, it hasn't. But it was failing with v1 end point. I might probably need to figure this out with @stevieing

tests/unit/stores/pacbioLibraryBatches.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      )
+    })
+  })
+
+  describe('actions', () => {
+    describe('createLibraryBatch', () => {
+      let create, getTagSets, csvFile, csvFileTextContent, logMessage, mockSuccessResponse, tagSet
+
+      beforeEach(() => {
+        vi.spyOn(apiGetter.traction.pacbio.requests, 'get').mockResolvedValue(
```
BenTopping, line 51: Do you need to spy on this? Can you do the same thing as below with the getTagSets?

tests/unit/stores/pacbioLibraryBatches.spec.js
```diff
@@ -0,0 +1,153 @@
...
+        expect(pacbioLibraryBatchStore.tubes).toEqual(pacbioLibraryBatchFactory.storeData.tubes)
+        expect(result).toEqual(pacbioLibraryBatchStore.librariesInBatch)
+      })
+
+      it('returns error when csv file contains duplicate tags', async () => {
+        csvFileTextContent = fs.readFileSync(
+          './tests/data/csv/pacbio_library_batch_duplicate_tags.csv',
+          'utf8',
+        )
+        create.mockResolvedValue(mockSuccessResponse)
```
BenTopping, line 115: Don't think this mockResolvedValue is needed

tests/unit/stores/pacbioLibraryBatches.spec.js
```diff
@@ -0,0 +1,153 @@
...
+        expect(success).toBeFalsy()
+        expect(errors).toEqual(['Duplicate tag: bc2067'])
+      })
+
+      it('returns error when csv file contains invalid source', async () => {
+        csvFileTextContent = fs.readFileSync(
+          './tests/data/csv/pacbio_library_batch_invalid_source.csv',
+          'utf8',
+        )
+        create.mockResolvedValue(mockSuccessResponse)
```
BenTopping, line 130: Again is this needed if its not being called?

src/stores/utilities/pacbioLibraryBatches.js
```diff
@@ -0,0 +1,108 @@
...
+    'volume',
+  ]
+  const requiredAttributes = ['source', 'tag', ...measurementAttributes]
+  for (const attr of requiredAttributes) {
+    if (!record[attr]) {
+      return createError(`${attr} is required`)
+    }
+  }
+
+  const request = requests.find((r) => r.source_identifier === source)
```
BenTopping, line 36: Handy to have a source_identifier 👍

src/stores/utilities/pacbioLibraryBatches.js
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Fetches all requests and tags for given tagset.
+ *
+ * @param {Object} tagSet - The tag set to fetch tags and requests for.
+ * @returns {Promise<Object>} - The fetched tags and requests.
+ */
+async function fetchTagsAndRequests(tagSet) {
+  const pacbioRootStore = usePacbioRootStore()
+  await store.dispatch('traction/pacbio/requests/setRequests')
+  const requests = store.getters['traction/pacbio/requests/requests']
```
BenTopping, line 102: I think the requests API is fairly performant but we will be overfetching data in the 1000s of records here. Probably doesn't matter at the moment but worth bearing in mind, maybe worth adding a TODO or thinking if there is another way to pull back the data.

tests/unit/stores/utilities/pabioLibraryBatches.spec.js
```diff
@@ -0,0 +1,146 @@
...
+    const requiredFields = [
+      'source',
+      'tag',
+      'concentration',
+      'insert_size',
+      'volume',
+      'template_prep_kit_box_barcode',
+    ]
+
+    requiredFields.forEach((field) => {
```
BenTopping, line 27: Nice.

BenTopping (changes requested):
Really nice, impressed with the error handling.

src/stores/utilities/pacbioLibraryBatches.js
seenanair: Good point. This is a question I’ve thought about as well. I understand that passing a large number of records at once isn’t ideal, but I was also concerned about making an API call for every single record. I’m happy to make changes if you think an alternative approach would be better.

tests/unit/stores/pacbioLibraryBatches.spec.js
seenanair: I think it is needed as we are checking the result against success.

BenTopping (approved):
Nice work 💯

tests/factories/PacbioLibraryBatchFactory.js
```diff
@@ -0,0 +1,192 @@
...
+          insert_size: library.insert_size,
+          template_prep_kit_box_barcode: library.template_prep_kit_box_barcode,
+        }
+      }),
+      tubes: tubesObj,
+      libraries: librariesObj,
+    }
+  }
+
+  const createLibraryBatchPayloadData = (csvText, requests) => {
```
stevieing, line 158: should this function be here. Are we mimicking what is going on elsewhere in which case we should be using the same functionality?

tests/factories/PacbioLibraryBatchFactory.js
```diff
@@ -0,0 +1,192 @@
...
+    ],
+    meta: {
+      page_count: 1,
+    },
+  }
+
+  const createStoreData = ({ included }) => {
+    const { tubes, libraries } = groupIncludedByResource(included)
+    const librariesObj = dataToObjectById({ data: libraries, includeRelationships: true })
+    const tubesObj = dataToObjectById({ data: tubes })
```
stevieing, line 140: same here. Not sure we should have functionality like this in a factory.

tests/data/csv/pacbio_library_batch.csv
```diff
@@ -0,0 +1,3 @@
+Source,Tag,Template prep kit box barcode,Volume,Concentration,Insert Size
```
stevieing, line 1: It would still be good to find a way we don't have to use csv files.

stevieing (approved):
Some advisory stuff in there.

In some ways putting complex logic in factories is another way to hide technical debt so it would be good to move it to external functions. We can look at this separately.

src/lib/csv/pacbio.js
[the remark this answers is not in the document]
seenanair: I didn't use reduce because here it makes an early return in case if any invalid records found which avoids parsing the rest . Hope it make sense.

tests/data/csv/pacbio_library_batch.csv
seenanair: I have now removed the CSV files, as they are now generated as part of the factory

stevieing (dismissed):

stevieing (approved):

PR #2078, opened 2024-11-28
BUG FIX - PacBio pool information

BenTopping:

- Fixes bug where updates to pool information commits the data to the service.

### DIFF
--- a/src/components/pacbio/PacbioPoolEdit.vue
+++ b/src/components/pacbio/PacbioPoolEdit.vue
@@ -230,7 +230,6 @@ const poolSetter = (attr) => {
     },
     set(newValue) {
       pool[attr] = newValue
-      updatePool(pool, attr)
     },
   })
 }
--- a/tests/unit/components/pacbio/PacbioPoolEdit.spec.js
+++ b/tests/unit/components/pacbio/PacbioPoolEdit.spec.js
@@ -80,28 +80,24 @@ describe('pacbioPoolEdit#new', () => {
       await input.setValue('017865101789500022821')
       await wrapper.vm.$nextTick()
       expect(store.pool.template_prep_kit_box_barcode).toEqual('017865101789500022821')
-      expect(store.updatePool).toBeCalled()
     })
 
     it('volume', async () => {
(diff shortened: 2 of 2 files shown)


src/stores/pacbioPoolCreate.js
```diff
@@ -1096,5 +1096,29 @@ export const usePacbioPoolCreateStore = defineStore('pacbioPoolCreate', {
...
+        pool.used_volume != null &&
+        parseFloat(pool[field]) < parseFloat(pool.used_volume)
+      ) {
+        pool.errors[field] = 'must be greater than used volume'
+        return
+      }
+
+      // Remove any existing errors for the field
+      delete pool.errors[field]
+    },
```
seenanair, line 1122: One doubt - I thought this added validation would be used to check if there are errors that prevent updating the pool, but I couldn’t see that in the updated code. Is this already implemented?

tests/unit/stores/pacbioPoolCreate.spec.js
```diff
@@ -1738,5 +1738,43 @@ describe('usePacbioPoolCreateStore', () => {
...
+          insert_size: 100,
+          template_prep_kit_box_barcode: 'ABC1',
+          errors: { concentration: ['error'] },
+        }
+        store.validatePoolAttribute('volume')
+        expect(store.pool.errors).toEqual({
+          concentration: ['error'],
+          volume: 'must be present',
+        })
+      })
```
seenanair, line 1777: May be one more test case with volume as 0 would be good.

src/stores/pacbioPoolCreate.js
BenTopping: Yeah its already implemented. There is a `validate` function called in updatePool and createPool that will double check the values before submitting to the service.

tests/unit/stores/pacbioPoolCreate.spec.js
BenTopping: I have updated the valid test to include volume as 0 and added a test for the used_volume check.

seenanair (approved):

PR #2080, opened 2024-11-29
update Reception page accordingly to the client feedback

sabrine33:
- Moved the workflow selector to the pipeline section.
- Filtered workflows based on the selected pipeline.
- Dynamically hide or show the user swipecard input depending on the workflow's status (selected or unselected).
- Added validation to the swipecard input: display an error message if it is not set and a workflow is selected.
- Displayed an error message if the Scan In operation fails due to missing required inputs.

I think we still need to update the swipecard fields (in different part of the app) to make them password fields, but I suppose this is can be addressed in different story


stevieing (approved):
Happy to approve. I presume this is working in local as expected?

sabrine33:
> Happy to approve. I presume this is working in local as expected?

yes, tested locally and it is working as expected

Issue #2008, opened 2024-10-21
Y24-413 - As a long read user (Michelle) I would like a page where I can upload a csv file to be able to create multiple libraries at once (24) to save time and reduce errors


stevieing:
**User story**
As a long read user (Michelle) I would like a page where I can upload a csv file to be able to create multiple libraries at once (24) to save time and reduce errors

**Who are the primary contacts for this story**
Michelle / Steve

**Who is the nominated tester for UAT**
Michelle / Long Read

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] There should be a page I can visit e.g. batch library creation
- [ ] I should be able to upload a csv in the attached format
- [ ] There should be an error when the csv is not in a valid format
- [ ] I should be able to select a tag set
- [ ] Once the csv is uploaded I should see a list of potential libraries with the associated data
- [ ] If there are any issues the the page shouls show errors e.g. invalid tage or sample, invalid data
- [ ] There should be a button allows me to create the libraries
- [ ] If there is an error then these should be returned and visible on the page
- [ ] If it is successful there should be a message
- [ ] I should be able to print the barcodes for the newly created libraries

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
[x24 traction library csv example.csv](https://github.com/user-attachments/files/17458820/x24.traction.library.csv.example.csv)
[637_ PacBio Library Preparation SPK3.0.xlsx](https://github.com/user-attachments/files/17458827/637_.PacBio.Library.Preparation.SPK3.0.xlsx)

PR #2084, opened 2024-12-03
Y24-413 library batch create page

seenanair:
Closes https://github.com/sanger/traction-ui/issues/2008

Page Component for Creating Library Batches:
-     Option to select a CSV file and tag set.
-     CSV preview display, which can be toggled on and off.
-     'Create Library Batch' button to initiate library batch creation.
-     Table to display created libraries.
-     Option to select a printer.
-     Barcode preview pane to display barcodes for printing.
-     Print button to print all barcodes.

Screenshot 2024-12-03 at 16 10 54

### DIFF
--- a/src/views/pacbio/PacbioLibraryBatchCreate.vue
+++ b/src/views/pacbio/PacbioLibraryBatchCreate.vue
@@ -1,0 +1,298 @@
+<template>
+  <DataFetcher :fetcher="fetchData">
+    <div class="w-full h-screen px-4">
+      <div class="flex flex-row w-full h-full">
+        <div class="w-1/2 px-4">
+          <div class="flex flex-col space-y-4">
+            <traction-section
+              title="Create Library Batch"
+              number="1"
+              description="To get started, please select a tag set and csv file, then press Create Libraries button"
+            ></traction-section>
+            <div class="text-left px-4">
+              Select tag set
+              <traction-select
+                v-model="selectedTagSet"
+                data-type="tag-set-list"
+                :options="pacbioRootStore.tagSetChoices"
+              ></traction-select>
+            </div>
+            <div class="px-4">
+              <label class="flex text-left" for="csvFieInput">Select file</label>
+              <div class="flex flex-row space-x-2">
+                <div id="borderDiv" class="w-full">
+                  <input
+                    id="csvFileInput"
+                    ref="csvFileInput"
+                    class="block rounded border file:border-0 w-full"
+                    type="file"
+                    accept="text/csv, .csv"
+                    @change="onSelectFile"
+                  />
+                </div>
+                <traction-button
+                  data-type="csv-preview-btn"
+                  class="whitespace-nowrap"
+                  :disabled="isDisabledCSVPreview"
+                  @click="showCSVPreview = !showCSVPreview"
+                  >{{ showCSVPreview ? 'Hide CSV Data' : 'Show CSV Data' }}</traction-button
+                >
+              </div>
+            </div>
+            <div v-if="showCSVPreview" class="flex flex-row space-x-4">
+              <div class="w-full h-64 overflow-y-auto border-2" data-type="csv-preview">
+                <traction-table :fields="state.csvTableFields" :items="state.csvData" />
+              </div>
+            </div>
+            <div class="flex flex-row space-x-8 py-4">
+              <traction-button
+                id="reset"
+                full-width
+                theme="reset"
+                data-action="reset-form"
+                @click="onReset"
+              >
+                Reset
+              </traction-button>
+              <traction-button
+                id="create"
+                :disabled="isCreateDisabled"
+                full-width
+                theme="create"
+                data-action="create-libraries"
+                @click="createLibraryBatch"
+              >
+                Create Libraries
+              </traction-button>
+            </div>
+
+            <traction-section
+              title="View Created Libraries"
+              number="2"
+              description="All libraries created will be displayed here"
+            ></traction-section>
+            <div class="flex flex-row space-x-4">
+              <div class="w-full overflow-y-auto">
+                <traction-table
+                  id="library-batch-table"
+                  :fields="state.resultTableFields"
+                  :items="state.resultData"
+                />
+                <div
+                  v-show="!state.resultData.length"
+                  data-type="created-libraries"
+                  class="text-sm font-bold text-black"
+                >
+                  No Libraries Created Yet
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+        <div class="w-1/2 px-4">
+          <div class="flex flex-col space-y-4">
+            <traction-section
+              title="Print Labels"
+              number="3"
+              description="Once the libraries are created, please select a printer and press the Print button"
+            ></traction-section>
+
+            <div class="text-left px-4 pb-4">
+              Select Printer
+              <traction-select
+                v-model="selectedPrinterName"
+                data-type="printer-options"
+                :options="printerOptions"
+              />
+            </div>
+
+            <div class="w-full h-full border-2">
+              <div class="w-full h-full p-4 space-y-4 rounded-md border-t-4">
+                <traction-heading level="4" show-border> Preview Barcodes </traction-heading>
+                <div class="h-full">
+                  <div class="space-x-4 pb-4 flex flex-row">
+                    <traction-button
+                      id="print-button"
+                      class="grow"
+                      :disabled="isPrintDisabled"
+                      theme="printRed"
+                      @click="onPrintAction"
+                    >
+                      Print Labels
+                    </traction-button>
+                  </div>
+
+                  <div tag="article" class="mb-2 text-black text-left">
+                    <div class="flex flex-col rounded p-4">
+                      <ul
+                        v-if="printBarcodes.length"
+                        id="list-barcodes-to-print"
+                        class="space-y-2 justify-center text-center"
+                      >
+                        <li
+                          v-for="barcode in printBarcodes"
+                          :key="barcode"
+                          class="text-sm text-sp font-bold"
+                        >
+                          {{ barcode }}
+                        </li>
+                      </ul>
+                      <div
+                        v-else
+                        data-type="empty-barcodes"
+                        class="justify-center text-center text-black text-sm font-bold"
+                      >
+                        No Barcodes to Print Yet
+                      </div>
+                    </div>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </DataFetcher>
+</template>
+
+<script setup>
+import { usePacbioRootStore } from '@/stores/pacbioRoot.js'
+import { usePrintingStore } from '@/stores/printing.js'
+import { usePacbioLibraryBatchCreateStore } from '@/stores/pacbioLibraryBatchCreate.js'
+import { computed, ref, reactive } from 'vue'
+import DataFetcher from '@/components/DataFetcher.vue'
+import { eachRecord } from '@/lib/csv/pacbio.js'
+import useAlert from '@/composables/useAlert.js'
+import usePacbioLibraryPrint from '@/composables/usePacbioLibraryPrint.js'
+
+const pacbioRootStore = usePacbioRootStore()
+const pacbioLibraryBatchCreateStore = usePacbioLibraryBatchCreateStore()
+const { printers, fetchPrinters } = usePrintingStore()
+const { showAlert } = useAlert()
+const { printLabels } = usePacbioLibraryPrint()
+
+const defaultTagSetName = 'Pacbio_96_barcode_plate_v3'
+
+const selectedPrinterName = ref('') // selected printer id
+const showCSVPreview = ref(false) // Show preview of the csv file
+const selectedTagSet = ref('') // Chosen tag set
+const csvFileInput = ref('') //Reference to the csv file input
+const selectedCSVFile = ref('') //Reference to the selected csv file
+
+/**
+ * Set the default tag set
+ *
+ */
+const setDefaultTagSet = () => {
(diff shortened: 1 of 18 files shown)


seenanair:
@stevieing @BenTopping It would be great if you could run this branch and check the page ui as part of the review.

src/views/pacbio/PacbioLibraryBatchCreate.vue
```diff
@@ -0,0 +1,302 @@
...
+    return
+  }
+  const { success, errors, result } = await pacbioLibraryBatchCreateStore.createLibraryBatch(
+    selectedCSVFile.value,
+    selectedTagSetName,
+  )
+  if (success) {
+    state.resultData = result
+  } else {
+    showAlert(errors, 'danger')
```
BenTopping, line 288: showAlert should take a string as an error not an array, its throwing a warning in the console. Might be worthing adding a join(',') to errors.

tests/e2e/specs/pacbio/pacbio_library_batch_create.cy.js
```diff
@@ -0,0 +1,132 @@
+import PacbioTagSetFactory from '../../../factories/PacbioTagSetFactory.js'
+import PrinterFactory from '../../../factories/PrinterFactory.js'
+import PacbioLibraryBatchFactory from '../../../factories/PacbioLibraryBatchFactory.js'
+import PacbioRequestFactory from '../../../factories/PacbioRequestFactory.js'
+const pacbioLibraryBatchFactory = PacbioLibraryBatchFactory()
+
+describe('Pacbio Pool Create', () => {
```
BenTopping, line 7: Pacbio Library batch

tests/e2e/specs/pacbio/pacbio_library_batch_create.cy.js
```diff
@@ -0,0 +1,132 @@
...
+
+    cy.get('@printerFactory').then((printerFactory) => {
+      cy.intercept('GET', '/v1/printers', {
+        statusCode: 200,
+        body: printerFactory.content,
+      })
+    })
+  })
+
+  it('Creates a pool successfully', () => {
```
BenTopping, line 35: Creates a library batch successfully.

src/views/pacbio/PacbioLibraryBatchCreate.vue
```diff
@@ -0,0 +1,302 @@
...
+const setDefaultTagSet = () => {
+  const defaultTagSet = pacbioRootStore.tagSetChoices.find(
+    (option) => option.text === defaultTagSetName,
+  )
+  if (defaultTagSet) {
+    selectedTagSet.value = defaultTagSet.value
+  }
+}
+
+async function fetchData() {
```
BenTopping, line 196: The page is unusable if the data fetch fails so not sure if this should always return true.

tests/factories/PacbioLibraryBatchFactory.js
```diff
@@ -1,5 +1,5 @@
 import BaseFactory from './BaseFactory'
-import { dataToObjectById, groupIncludedByResource } from '@/api/JsonApi'
+import { dataToObjectById, groupIncludedByResource } from './../../src/api/JsonApi'
```
BenTopping, line 2: Does @ not work?

tests/factories/PacbioRequestFactory.js
```diff
@@ -1,5 +1,5 @@
 import BaseFactory from './BaseFactory.js'
-import { dataToObjectById } from '@/api/JsonApi.js'
+import { dataToObjectById } from './../../src/api/JsonApi'
```
BenTopping, line 2: Same as above.

tests/unit/composables/usePacbioLibraryPrint.spec.js
```diff
@@ -0,0 +1,56 @@
...
+      fourth_line: 'SQSC-2',
+      label_name: 'main_label',
+    },
+  ]
+  beforeEach(() => {
+    const pinia = createPinia()
+    setActivePinia(pinia)
+  })
+
+  it('prints labels correctly', async () => {
```
BenTopping, line 39: test for failure

BenTopping (changes requested):
Works well, nice validation and clear error messages.

Some notes while using:
- I think the csv data should always be visible as they will want to confirm what they have uploaded is not missing any data so I am not sure about the show/hide csv data button. If its required to keep I would have it set to be showing by default.
- You can press 'create libraries' multiple times and create several sets of the same libraries without changing the input file.
- Small UX note, The preview barcodes has a slightly odd top border and I think 'print labels' button should be below the list of barcodes.

seenanair:
Thanks for the comments Ben.
>>I think the csv data should always be visible as they will want to confirm what they have uploaded is not missing any data so I am not sure about the show/hide csv data button. If its required to keep I would have it set to be showing by default
The reasoning behind the Show/Hide button is that we do not have a limit on the number of records in the CSV. As a result, the table size can take up significant screen space (, which may reduce the viewport available for viewing the "Created Libraries" table. I thought the "Created Libraries" table would be more important to prioritize.
I will combine this into a single table. When a CSV is selected, that section will display "Preview CSV." Once the user creates libraries, it will update to "View Libraries," with the data reflecting the returned barcodes. I hope that makes sense.
>>You can press 'create libraries' multiple times and create several sets of the same libraries without changing the input file.
Do we need to prevent that? If so, "Create Libraries" can be disabled once the libraries are successfully created by resetting the CSV file input field.
>>Small UX note, The preview barcodes has a slightly odd top border and I think 'print labels' button should be below the list of barcodes.
Ok. In a similar scenario on the Label Printing page, we have a "Print" button at the top. I thought it would be good to maintain consistency, especially since there was feedback in the past about displaying the button at the top to avoid users having to scroll to press it if the list is too long. I'm happy either way.


> Works well, nice validation and clear error messages.
> 
> Some notes while using:
> 
> * I think the csv data should always be visible as they will want to confirm what they have uploaded is not missing any data so I am not sure about the show/hide csv data button. If its required to keep I would have it set to be showing by default.
> * You can press 'create libraries' multiple times and create several sets of the same libraries without changing the input file.
> * Small UX note, The preview barcodes has a slightly odd top border and I think 'print labels' button should be below the list of barcodes.

tests/factories/PacbioLibraryBatchFactory.js
seenanair: No, it won't work when factories are used in e2e tests. So we have used it this way across all factories.

src/views/pacbio/PacbioLibraryBatchCreate.vue
seenanair: Good point, will change it to displaying an error message in case of failure

tests/factories/PacbioLibraryBatchFactory.js
stevieing: The @ is a vite / vue alias so is not available in the browser.

src/views/pacbio/PacbioLibraryBatchCreate.vue
seenanair: I think addMessage in rootStore is expecting errors as an object. So, I will change the store where it is returned as an array

src/views/pacbio/PacbioLibraryBatchCreate.vue
```diff
@@ -0,0 +1,307 @@
...
+  }
+}
+
+/**
+ * Prints the labels for the created libraries
+ */
+const onPrintAction = async () => {
+  const { success, message = {} } = await printLabels(
+    selectedPrinterName.value,
+    state.resultData.map((item) => ({ barcode: item.barcode, source_identifier: item.source })),
```
stevieing, line 303: Would this be better as a getter?

tests/e2e/specs/pacbio/pacbio_library_batch_create.cy.js
```diff
@@ -0,0 +1,130 @@
...
+    })
+  })
+
+  it('Creates a library batch successfully', () => {
+    cy.visit('#/pacbio/library-batch')
+    cy.contains('Create Library Batch')
+
+    // Upload CSV file
+    //Create a file from PacbioLibraryBatchFactory and attach it to the input
+    const csvContent = pacbioLibraryBatchFactory.createCsvFromLibraryBatchData(
```
stevieing, line 41: Nice. Like the fact that we now have a central way of creating the CSV data.

tests/factories/PacbioTagSetFactory.js
```diff
@@ -23,6 +23,7 @@ const createStoreData = (data) => {
         first: (n = 1) => tagSet.tags.slice(0, n).map((id) => tags[id]),
       },
     },
+    findTagSetByName: (name) => Object.values(tagSets).find((ts) => ts.name === name),
```
stevieing, line 26: These are great but wondering if we are doing this in the actual code and if so should it be a function elsewhere?

stevieing (changes requested):
Pretty good. The factories have maded things much easier.

Just a few things around unneeded expectations.

Generally testing whether something is there or is disable is checking implementation rather than behaviour.

I will try it locally.

seenanair:
Did it still create libraries with invalid samples?
> When I upload the batch and the samples are not valid I get [no requests found]. It might be worth asking Michelle to upload a batch with invalid samples to see what she says about this error message?
> 
> When I uploaded the batch the source was not show Screenshot 2024-12-04 at 14 34 09 n.

stevieing:
> Did it still create libraries with invalid samples? When I tried it didn't create any libraries and gave this error message "Invalid record at line 2: source GEN-1725896371-4:B300 not found"
> 
> > When I upload the batch and the samples are not valid I get [no requests found]. It might be worth asking Michelle to upload a batch with invalid samples to see what she says about this error message?
> > When I uploaded the batch the source was not show Screenshot 2024-12-04 at 14 34 09 n.

It created libraries with valid samples. The sources are there 
Screenshot 2024-12-04 at 15 03 37

tests/factories/PacbioTagSetFactory.js
seenanair: I don't think we currently have a similar function elsewhere.

tests/factories/PrinterFactory.js
```diff
@@ -15,6 +15,8 @@ const createStoreData = (data) => {
     selected: {
       printer: printers['1'],
     },
+    getPrintersOfType: (type) =>
```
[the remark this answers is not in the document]
seenanair: 👍 A similar thing was used in store, so I have moved this to a utility file so it can be used across test as well.

seenanair:
Good finding 👍 . The missing sources was a bug, which got fixed now

> > Did it still create libraries with invalid samples? When I tried it didn't create any libraries and gave this error message "Invalid record at line 2: source GEN-1725896371-4:B300 not found"
> > > When I upload the batch and the samples are not valid I get [no requests found]. It might be worth asking Michelle to upload a batch with invalid samples to see what she says about this error message?
> > > When I uploaded the batch the source was not show Screenshot 2024-12-04 at 14 34 09 n.
> 
> It created libraries with valid samples. The sources are there Screenshot 2024-12-04 at 15 03 37

tests/unit/views/pacbio/PacbioLibraryBatchCreate.spec.js
```diff
@@ -0,0 +1,356 @@
...
+      )
+
+      const csvInput = wrapper.find('input[id="csvFileInput"]')
+      expect(csvInput.exists()).toBe(true)
+      const csvPreviewBtn = wrapper.find('button[data-type="csv-preview-btn"]')
+      expect(csvPreviewBtn.exists()).toBe(true)
+      expect(csvPreviewBtn.element.disabled).toBe(true)
+      expect(wrapper.find('div[data-type="csv-preview"]').exists()).toBe(false)
+
+      const resetBtn = wrapper.find('#reset')
```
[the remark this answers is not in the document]
seenanair: > Advisory: I think generally testing whether a button is there and disabled can lead to flaky tests as you are not testing behaviour. Generally better to test what happens if you press it in which case it might already be covered by e2e tests.
I’ve added a bunch of tests for enabling and disabling checks because this page goes through a lot of state changes depending on different actions. These changes affect the button states, so I thought it as more of a behavior test in which case unit tests are the best fit for this since they’re less flaky than e2e tests. Also, then we don’t need to cover all of it in e2e tests, which keeps things consistent with how other pages are set up—unit tests for detailed behaviors and e2e tests for the bigger picture. Hope it makes sense.

BenTopping (approved):
Looks good, nice work.

src/views/pacbio/PacbioLibraryBatchCreate.vue
```diff
@@ -0,0 +1,312 @@
...
+import useAlert from '@/composables/useAlert.js'
+import usePacbioLibraryPrint from '@/composables/usePacbioLibraryPrint.js'
+
+const pacbioRootStore = usePacbioRootStore()
+const pacbioLibraryBatchCreateStore = usePacbioLibraryBatchCreateStore()
+const { printers, fetchPrinters } = usePrintingStore()
+const { showAlert } = useAlert()
+const { printLabels } = usePacbioLibraryPrint()
+
+const defaultTagSetName = 'Pacbio_96_barcode_plate_v3'
```
stevieing, line 177: Something for later but it is likely that the default tag set will change so it might be worth adding this as a flag in the tag set table. Maybe we need a story to add this to the service.

tests/unit/views/pacbio/PacbioLibraryBatchCreate.spec.js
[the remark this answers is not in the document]
stevieing: Ok. It might be worth adding this as a composable so we can dynamically define the state and state changes and then have centralise the tests for it. Maybe worth a technical debt story?

stevieing (approved):
I have approved. Good work.

Couple of things for new stories.

PR #2107, opened 2024-12-11
Y24-354 - Speed up ONT Run page

### LINKED ISSUE #1984: Y24-354 - As developer we want to improve the ONT runs page is it currently runs very slowly.
**User story**
As developer we want to improve the ONT runs page is it currently runs very slowly. When the visit the runs/new page or edit an existing run it takes ~18 seconds for it to load.

This is because we are pulling back all of the pools for each run. It will only keep getting slower as the number of pools increases.

Steve / Seena

PSD / Long Read

**Acceptance criteria**
- [ ] The runs/new page should open in less than a second
- [ ] When I edit an existing run the page should open in less than a second
- [ ] There should be an end to end test to ensure the page is working correctly.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1984 

- Changes ONT Flowcell validation.
- Changes ONT Flowcell pool barcode check to be on demand instead of fetching all pool barcodes on page load.
- ONT run edit/creation page load speed ⚡ 


src/stores/ontRuns.js
```diff
@@ -97,13 +106,37 @@ export const useOntRunsStore = defineStore('ontRuns', {
...
-        const existingPools = store.getters['traction/ont/pools/pools']
+        const { pools } = groupIncludedByResource(included)
+        this.pools = formatById(this.pools, pools, true)
 
-        this.currentRun = buildFormatedOntRun(existingInstruments, existingPools, data, included)
+        this.currentRun = buildFormatedOntRun(existingInstruments, this.pools, data, included)
         return { success, errors }
       }
     },
+    async fetchPool(barcode) {
```
stevieing, line 116: This could do with some jsdoc comments.

stevieing (changes requested):
All looks good. Just worth adding some documentation where specified.

I have not tested the performance improvements but will have a look when it gets to UAT.

seenanair (commented):
Looking good.
Just one thought - As ONTFlowcell is refactored now, does it make sense to change that to composition api  as well?

src/stores/ontRuns.js
```diff
@@ -97,13 +106,45 @@ export const useOntRunsStore = defineStore('ontRuns', {
...
+        }
+      }
+      const rootStore = useRootStore()
+      const request = rootStore.api.traction.ont.pools
+      const promise = request.get({ filter: { barcode } })
+      const response = await handleResponse(promise)
+      let { success, body: { data } = {} } = response
+
+      // If response is successful and the
+      if (success && data.length > 0) {
```
stevieing, line 138: I see this all over the place it makes me feel that we should add another flag to the response handler to check that there is data.

Don't think we need to do it here but maybe put a note on it.

src/views/ont/ONTRun.vue
```diff
@@ -75,10 +75,14 @@ export default {
       return this.actions[this.newRecord ? 'create' : 'update']
     },
     runValid() {
-      const flowcells = (this.currentRun.flowcell_attributes || []).filter(
-        (fc) => fc.flowcell_id && fc.tube_barcode,
-      )
-      return this.currentRun.instrument_name && this.currentRun.state && flowcells.length != 0
+      const flowCellsValid = this.currentRun.flowcell_attributes?.every((fc) => {
```
stevieing, line 78: Advisory. Something for another time. Feel like this belongs elsewhere.

src/components/ont/runs/ONTFlowcell.vue
```diff
@@ -85,110 +83,73 @@ export default {
...
-    },
-    barcodeValidationError() {
-      return this.flowCellValidationState.errorBarcode
-    },
-    isBarcodeExists() {
-      return this.flowCellValidationState.statusBarcode !== FieldStatusEnum.Empty
+      return this.getFlowCell(this.position) ?? { ...flowCellType() }
     },
     /**Displays green if valid, red if invalid and no border if empty */
     flowcell_id_field_colour() {
```
stevieing, line 114: I was sure Vue had a really clever way of managing dynamic classes but not sure how it would work here.

stevieing (approved):
Looking good.

A couple of advisories for a future date.

src/components/ont/runs/ONTFlowcell.vue
BenTopping: Some interesting stuff in https://vuejs.org/guide/essentials/class-and-style. Not sure how much of an improvement it is though.

stevieing (approved):
Some nice refactors to make things more explicit and remove some unneeded store usage.

seenanair (approved):
Looking good. Nice refactoring 👍

stevieing (approved):

PR #2121, opened 2024-12-16
Y24-149 exhaust samples moved to destroyed location

seenanair:
Closes #

- Exhaust library volumes for all barcodes scanned to destroyed location


src/services/labwhere/client.js
```diff
@@ -89,5 +95,52 @@ const scanBarcodesInLabwhereLocation = async (
...
+ * @param {Array<string>} labwareBarcodes - The barcodes of the labware.
+ * @returns {Promise<Object>} - An object containing the success status and the exhausted libraries.
+ */
+const exhaustLibraryVolumeIfDestroyed = async (locationBarcode, labwareBarcodes) => {
+  if (locationBarcode !== destroyLocation) return { success: false }
+  let librariesToDestroy = []
+
+  //Fetch libraries by filter key
+  const fetchAndMergeLibraries = async (barcodes, filterKey) => {
+    const filterOptions = { filter: { [filterKey]: barcodes.join(',') } }
```
BenTopping, line 111: Did you try included both source identifier and barcode filters at the same time? Not sure if that only returns records matching one or both of the filters.

src/stores/pacbioLibraries.js
```diff
@@ -155,47 +121,28 @@ export const usePacbioLibrariesStore = defineStore('pacbioLibraries', {
...
-        filter,
-        include: 'request,tag,tube',
-      })
-      const response = await handleResponse(promise)
-
-      const { success, body: { data, included = [], meta = {} } = {}, errors = [] } = response
-
+    async fetchLibraries(filterOptions) {
+      const { success, data, meta, errors, libraries, tubes, tags, requests } =
+        await fetchLibraries({
```
BenTopping, line 126: I think this naming should be changed so its clear this isnt a recursive function.

src/stores/utilities/pacbioLibraries.js
BenTopping, line 1: What decides whether a method goes in utilities or goes into the store? I normally consider utilities as data formatting, etc but some of these make requests to the service as well. Not sure whether that belongs better in the store or not.

src/views/LabwhereReception.vue
```diff
@@ -170,7 +173,17 @@ const scanBarcodesToLabwhere = async () => {
...
     )
     if (response.success) {
-      showAlert(response.message, 'success')
+      let message = response.message
+      // Check if the library volume need to be exhausted
+      const { success, exhaustedLibraries } = await exhaustLibraryVolumeIfDestroyed(
+        location_barcode.value,
+        uniqueBarcodesArray.value,
+      )
+      if (success && exhaustedLibraries.length > 0) {
```
BenTopping, line 182: This message formatting should probably be handled by `exhaustLibraryVolumeIfDestroyed` as exhaustedLibraries are only used for the message.

BenTopping (commented):
Good start, couple questions.

Do we have some way in Traction UI of representing exhausted libraries. Is an exhausted library just one thats volume equals its used_volume or does it also have to be destroyed? Might be useful to have an exhausted badge or something?

seenanair:
> Good start, couple questions.
> 
> Do we have some way in Traction UI of representing exhausted libraries. Is an exhausted library just one thats volume equals its used_volume or does it also have to be destroyed? Might be useful to have an exhausted badge or something?

I thought the same.
At present, there is no representation of exhausted libraries.
 It would be good 
- to have an indication to show whether a library is exhausted and 
- to disable or doesn't allow editing the volume for exhausted libraries, as it doesn’t make sense to allow it again.

src/services/labwhere/client.js
seenanair: Yes, I tried. but it only returned the ones matching both filters like an AND operator, which won't give the intended results for this case.

stevieing (dismissed):

.env.development
```diff
@@ -2,7 +2,8 @@ VITE_TRACTION_BASE_URL=http://traction
...
 VITE_PRINTMYBARCODE_BASE_URL=http://printmybarcode
 VITE_SAMPLEEXTRACTION_BASE_URL=http://sampleextraction
-VITE_LABWHERE_BASE_URL=http://labwhere
+VITE_LABWHERE_BASE_URL=http://localhost:3003
 VITE_SEQUENCESCAPE_API_KEY=development
 VITE_LOG=true
-VITE_ENVIRONMENT=development
\ No newline at end of file
+VITE_ENVIRONMENT=development
+VITE_DESTROYED_LOCATION_BARCODE=lw-destroyed-217
```
stevieing, line 9: Are youn planning to add the destroyed location to the deployment project or just the .env files?

.env.development
```diff
@@ -2,7 +2,8 @@ VITE_TRACTION_BASE_URL=http://traction
 VITE_TRACTION_GRAPHQL_URL=http://traction/graphql
 VITE_PRINTMYBARCODE_BASE_URL=http://printmybarcode
 VITE_SAMPLEEXTRACTION_BASE_URL=http://sampleextraction
-VITE_LABWHERE_BASE_URL=http://labwhere
+VITE_LABWHERE_BASE_URL=http://localhost:3003
```
stevieing, line 5: I think it is better to do this in .env.development.local

src/services/labwhere/client.js
```diff
@@ -1,7 +1,13 @@
 import { extractLocationsForLabwares } from './helpers.js'
 import { FetchWrapper } from '@/api/FetchWrapper.js'
+import {
+  exhaustLibrayVolume,
+  formatAndTransformLibraries,
+} from '@/stores/utilities/pacbioLibraries.js'
+import { getPacbioLibraryResources } from '@/services/traction/PacbioLibrary.js'
 
 const labwhereFetch = FetchWrapper(import.meta.env['VITE_LABWHERE_BASE_URL'], 'LabWhere')
```
stevieing, line 9: **Advisory**: Creating the variables like this rather than using dependency injection makes it more difficult to test which is why you need to mock fetch globally I think!

A different way to do it would be to pass a type into the function which could have the various bits of configuration.

tests/unit/services/labwhere/client.spec.js
```diff
@@ -1,130 +1,240 @@
...
+import * as pacbioLibraryService from '@/services/traction/PacbioLibrary.js'
+import { createPinia, setActivePinia } from '@support/testHelper.js'
+import { beforeEach, describe, it } from 'vitest'
+const mockFetch = vi.fn()
+const labwhereLocationsFactory = LabwhereLocationsFactory()
 
-describe('client', () => {
-  const mockFetch = vi.fn()
-  const labwhereLocationsFactory = LabwhereLocationsFactory()
+beforeEach(() => {
```
stevieing, line 14: This is what I am referring to above.

We are essentially mocking fetch globally because of what we are doing in the client.

tests/unit/services/labwhere/client.spec.js
```diff
@@ -1,130 +1,240 @@
...
+      requests: {},
+    },
+  ]
+  const formattedLibraries = [
+    { id: 1, source_identifier: 'barcode1', barcode: 'TRAC-1' },
+    { id: 2, source_identifier: 'barcode2', barcode: 'TRAC-2' },
+  ]
+
+  beforeEach(() => {
+    vi.clearAllMocks()
```
stevieing, line 155: Not entirely sure what is going on but would be slightly concerned that we are pretty much mocking all of the functions. It would be good to know why. This could create technical debt.

stevieing (changes requested):
I have added lots of advisory comments.

I struggled a bit to understand what was going on as there is a lot of mocking.

I don't think these comments should stop deployment from branch if it is working.

.env.development
seenanair: It is already added in deployment project https://github.com/sanger/deployment/pull/533

tests/unit/services/labwhere/client.spec.js
seenanair: The ones you mentioned are not included as part of this PR.
The current implementation of the LabWhere client methods and FetchWrapper directly calls the fetch API to interact with LabWhere services. I believe this approach was taken because the LabWhere APIs do not always return JSON responses, and the client methods are designed as wrappers around these APIs to handle various response formats and content types.
These methods are globally mocked since all exported methods involve calling LabWhere, and the tests need to simulate different response formats for different methods. As you mentioned, it could likely be rewritten to pass in a fetch function, allowing tests to be written without relying on global mocking.

.env.development
seenanair: There is an e2e test to check for volume exhaustion when the entered location barcode matches the destroyed location barcode, as specified in the corresponding .env file. This test will fail during CI if the correct configuration is not set. Is there are any better approaches to handle this scenario?

stevieing: Ah ok. That is a problem. How does it work for all of our other api calls. We don't set traction url to a real one. It sounds like the intercept is not being handled correctly maybe?

seenanair: Here, we are simply comparing the entered barcode with the value of the environment variable read within the method. I think this is different from other cases where we can intercept commands to mock network requests and responses, eliminating the need for the real ones. I tried various approaches, such as using Cypress config to define a value, but it didn't work

stevieing (approved):
Nice piece of work.

I think we could also use dependency injection for the destroyed location barcode and turn it into a type but I think you have done more than enough on this for now. There will be more work on location tracking so we can look again in the future.

PR #2163, opened 2025-01-28
Y24-500 - Fixes existing pool to new pool navigation data integrity problems

### LINKED ISSUE #2096: Y24-500 - [BUG] - When you move from editing a pool to a new pool the data persists. This could cause confusion and data integrity issues
**Describe the bug**
When you move from editing a pool to a new pool the data persists. This could cause confusion and data integrity issues

**RT Ticket Number**
If applicable

**To Reproduce**
1. Go to page '...'
2. Click on button '...'
3. See error

**Expected behaviour**
What was expected to happen.

**Screenshots**

* Windows or Mac?

**Additional context**
It looks like we need to clear the data when we move to the new pool page for Pacbio and ONT.

BenTopping:
Closes #2096 

- Add key to data fetcher component to force rerender when key (route) changes.
- Add resetData function to PacbioPoolCreate to reset data when rerendering the component.

### DIFF
--- a/src/views/ont/ONTPoolCreate.vue
+++ b/src/views/ont/ONTPoolCreate.vue
@@ -1,5 +1,6 @@
 <template>
-  <DataFetcher id="pool" :fetcher="provider">
+  <!-- The data fetcher key is used to re-render the page if a user goes from an existing pool to a new one -->
+  <DataFetcher id="pool" :key="$route.fullPath" :fetcher="provider">
     <div class="flex flex-col pt-4">
       <div class="flex flex-col w-1/2 px-4">
         <traction-menu :border="true">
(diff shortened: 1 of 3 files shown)


seenanair (commented):
Looking good 👍 
Do you think we need an e2e test to cover the scenario of transitioning from the 'Edit Pool' page to the 'Pool/New' page?

BenTopping:
> Looking good 👍 Do you think we need an e2e test to cover the scenario of transitioning from the 'Edit Pool' page to the 'Pool/New' page?

Thanks. I don't think an e2e test is needed as this is an uncommon use case, what do you think?

seenanair:
> > Looking good 👍 Do you think we need an e2e test to cover the scenario of transitioning from the 'Edit Pool' page to the 'Pool/New' page?
> 
> Thanks. I don't think an e2e test is needed as this is an uncommon use case, what do you think?
Yes, I agree. That’s a very specific case, so it may not be necessary. Our end-to-end tests typically don’t cover such detailed scenarios.

seenanair (approved):

PR #2206, opened 2025-02-17
2173-y25-084_include available volume in the libraries page

wendyyang:
Closes #

##### add 'available volume' to library table
- 

### DIFF
--- a/src/components/pacbio/PacbioLibraryEdit.vue
+++ b/src/components/pacbio/PacbioLibraryEdit.vue
@@ -90,7 +90,7 @@ const updateLibrary = async () => {
     formRef.value.formLibrary,
   )
   if (success) {
-    emit('editCompleted', updatedLibrary) // Emits a 'editCompleted' event which the parent component can listen to.
+    emit('editCompleted') // Emits a 'editCompleted' event which the parent component can listen to.
     showAlert('Updated library with barcode ' + props.library.barcode, 'success')
   } else {
     showAlert(`Failed to update library in Traction: ${errors}`, 'danger')
--- a/src/components/pacbio/PacbioLibraryEdit.vue
+++ b/src/components/pacbio/PacbioLibraryEdit.vue
@@ -86,9 +86,11 @@ const librariesStore = usePacbioLibrariesStore()
  * If the update is successful, it shows a success message, otherwise shows a failure message.
  */
 const updateLibrary = async () => {
-  const { success, errors } = await librariesStore.updateLibrary(formRef.value.formLibrary)
+  const { success, errors, updatedLibrary } = await librariesStore.updateLibrary(
+    formRef.value.formLibrary,
+  )
   if (success) {
-    emit('editCompleted') // Emits a 'editCompleted' event which the parent component can listen to.
+    emit('editCompleted', updatedLibrary) // Emits a 'editCompleted' event which the parent component can listen to.
     showAlert('Updated library with barcode ' + props.library.barcode, 'success')
   } else {
     showAlert(`Failed to update library in Traction: ${errors}`, 'danger')
--- a/src/components/pacbio/PacbioLibraryEdit.vue
+++ b/src/components/pacbio/PacbioLibraryEdit.vue
@@ -86,9 +86,7 @@ const librariesStore = usePacbioLibrariesStore()
  * If the update is successful, it shows a success message, otherwise shows a failure message.
  */
 const updateLibrary = async () => {
-  const { success, errors } = await librariesStore.updateLibrary(
-    formRef.value.formLibrary,
-  )
+  const { success, errors } = await librariesStore.updateLibrary(formRef.value.formLibrary)
   if (success) {
     emit('editCompleted') // Emits a 'editCompleted' event which the parent component can listen to.
     showAlert('Updated library with barcode ' + props.library.barcode, 'success')
--- a/src/components/pacbio/PacbioLibraryEdit.vue
+++ b/src/components/pacbio/PacbioLibraryEdit.vue
@@ -86,7 +86,7 @@ const librariesStore = usePacbioLibrariesStore()
  * If the update is successful, it shows a success message, otherwise shows a failure message.
  */
 const updateLibrary = async () => {
-  const { success, errors, updatedLibrary } = await librariesStore.updateLibrary(
+  const { success, errors } = await librariesStore.updateLibrary(
     formRef.value.formLibrary,
   )
   if (success) {
--- a/src/services/traction/PacbioLibrary.js
+++ b/src/services/traction/PacbioLibrary.js
@@ -96,8 +96,8 @@ const buildLibraryResourcePayload = ({
 async function updatePacbioLibraryResource(libraryFields) {
   const request = useRootStore().api.traction.pacbio.libraries
   const promise = request.update(buildLibraryResourcePayload(libraryFields))
-  const response = await handleResponse(promise)
-  return { success: response.success, errors: response.errors, updatedLibrary: response.body }
+  const { success, body: { data: { attributes: updatedLibrary } = {} } = {}, errors } = await handleResponse(promise)
+  return { success, errors, updatedLibrary }
 }
 
 export { getPacbioLibraryResources, updatePacbioLibraryResource, buildLibraryResourcePayload }
--- a/src/services/traction/PacbioLibrary.js
+++ b/src/services/traction/PacbioLibrary.js
@@ -96,7 +96,11 @@ const buildLibraryResourcePayload = ({
 async function updatePacbioLibraryResource(libraryFields) {
   const request = useRootStore().api.traction.pacbio.libraries
   const promise = request.update(buildLibraryResourcePayload(libraryFields))
-  const { success, body: { data: { attributes: updatedLibrary } = {} } = {}, errors } = await handleResponse(promise)
+  const {
+    success,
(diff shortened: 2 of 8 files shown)


src/views/pacbio/PacbioLibraryIndex.vue
```diff
@@ -268,6 +269,11 @@ const getEditLibrary = (row) => {
   return value
 }
 
+// Add this function to refetch after editing
+const handleEditComplete = async () => {
```
dasunpubudumal, line 273: Small question: I tried in UAT by editing a library (I tried updating the initial volume). I submitted the form after I edited the library, and it didn't have to reload the libraries for the change to be updated in the libraries UI table. Is this `handleEditComplete` function doing something different?

seenanair: With this PR, we're now showing the 'Available Volume' column in table, which should update accordingly when the Initial Volume changes. However, right now, the Initial Volume field updates using the given value from the UI after a successful update call, but the Available Volume doesn’t reflect the change. 
Ideally, we should update the library object with the response from the server so all fields stay in sync.

dasunpubudumal: Ah alright. Got it.

src/stores/pacbioLibraries.js
```diff
@@ -144,15 +144,20 @@ export const usePacbioLibrariesStore = defineStore('pacbioLibraries', {
...
+
       if (success) {
-        //Update all fields of the library in the store with matching ID with the given values.
-        this.libraries[libraryFields.id] = {
-          ...this.libraries[libraryFields.id],
-          ...libraryFields,
-        }
+        // Update the library in the store if the server update was successful
+
+        this.$patch((state) => {
```
BenTopping, line 152: You shouldn't need `this.$patch`. It should just work as before using `this.libraries`

tests/e2e/specs/pacbio/pacbio_libraries_view.cy.js
```diff
@@ -86,14 +97,15 @@ describe('Pacbio Libraries view', () => {
...
+    cy.get('#available_volume').first().should('have.text', '3')
     cy.get('#concentration').first().should('have.text', '2')
     cy.get('#template_prep_kit_box_barcode').first().should('have.text', 'LK54321')
     cy.get('#insert_size').first().should('have.text', '200')
-    cy.get('@pacbioTagSetFactory').then((pacbioTagSetFactory) => {
-      cy.get('#tag_group_id')
-        .first()
-        .should('have.text', pacbioTagSetFactory.storeData.selected.tag.group_id)
-    })
+    // cy.get('@pacbioTagSetFactory').then((pacbioTagSetFactory) => {
```
BenTopping, line 105: Can we uncomment this?

tests/e2e/specs/pacbio/pacbio_libraries_view.cy.js
```diff
@@ -86,14 +97,15 @@ describe('Pacbio Libraries view', () => {
     cy.get('#libraryForm').should('not.exist')
 
     //It should display updated values in table
-    cy.get('#volume').first().should('have.text', '2')
+    cy.get('#volume').first().should('have.text', '3')
+    cy.get('#available_volume').first().should('have.text', '3')
```
BenTopping, line 101: Can we also check the available_volume has changed by checking its original value above - around line 79.

src/stores/utilities/pacbioLibraries.js
```diff
@@ -87,8 +87,13 @@ async function validateAndUpdateLibrary(libraryFields) {
   if (!valid.success) {
     return valid
   }
-  const { success, errors } = await updatePacbioLibraryResource(libraryFields)
-  return { success, errors }
+  // const { success, errors, updatedLibrary } = await updatePacbioLibraryResource(libraryFields)
+  const {
+    success,
+    updatedLibrary: { data, included = [] } = {},
```
BenTopping, line 93: Can we update the tests in tests/unit/stores/utilities/pacbioLibraries.spec.js to also check the behaviour of this updateLibrary field.

src/services/traction/PacbioLibrary.js
```diff
@@ -96,8 +96,8 @@ const buildLibraryResourcePayload = ({
 async function updatePacbioLibraryResource(libraryFields) {
   const request = useRootStore().api.traction.pacbio.libraries
   const promise = request.update(buildLibraryResourcePayload(libraryFields))
-  const { success, errors } = await handleResponse(promise)
-  return { success, errors }
+  const response = await handleResponse(promise)
```
BenTopping, line 99: Advisory: In some places we used the spread syntax here to extract the values. E.g. `const { success, errors, body: {} } = handleResponse(promise)`. This makes the return block below slightly less verbose.

BenTopping (commented):
Looking good, think it is almost there.

src/components/pacbio/PacbioLibraryEdit.vue
```diff
@@ -86,9 +86,11 @@ const librariesStore = usePacbioLibrariesStore()
  * If the update is successful, it shows a success message, otherwise shows a failure message.
  */
 const updateLibrary = async () => {
-  const { success, errors } = await librariesStore.updateLibrary(formRef.value.formLibrary)
+  const { success, errors, updatedLibrary } = await librariesStore.updateLibrary(
+    formRef.value.formLibrary,
+  )
   if (success) {
-    emit('editCompleted') // Emits a 'editCompleted' event which the parent component can listen to.
+    emit('editCompleted', updatedLibrary) // Emits a 'editCompleted' event which the parent component can listen to.
```
seenanair, line 93: I don't think you need to emit the updatedLibrary as the parent component doesn't do anything with it.

src/services/traction/PacbioLibrary.js
seenanair: I agree with Ben's comment. I’d suggest using:
  `const { success, body: { data: { attributes: updatedLibrary } = {} } = {}, errors } = await handleResponse(promise)` 
This way, all the destructuring is in one place, keeping it simple and clean. You can then just return updatedLibrary.

src/stores/utilities/pacbioLibraries.js
seenanair: As I mentioned in the comment above, if we do the destructuring in the updatePacbioLibraryResource method where the response is handled, we can avoid the need for destructuring here.

src/stores/utilities/pacbioLibraries.js
```diff
@@ -87,8 +87,13 @@ async function validateAndUpdateLibrary(libraryFields) {
   if (!valid.success) {
     return valid
   }
-  const { success, errors } = await updatePacbioLibraryResource(libraryFields)
-  return { success, errors }
+  // const { success, errors, updatedLibrary } = await updatePacbioLibraryResource(libraryFields)
```
seenanair, line 90: Would you please remove the commented code.

tests/unit/stores/pacbioLibraries.spec.js
```diff
@@ -227,8 +229,23 @@ describe('usePacbioLibrariesStore', () => {
...
+        const mockResponse = {
+          status: 201,
+          json: async () => ({
+            data: {
+              id: library.id,
+              attributes: {
+                concentration: 2.0,
+                template_prep_kit_box_barcode: 'LK12348',
+                volume: 4.0,
+                tag_id: '3',
```
seenanair, line 241: Maybe it is good to add available_volume as well to these attributes?

seenanair (commented):
Good work, it's getting there. Just mentioned a few thoughts.

src/stores/utilities/pacbioLibraries.js
wendyyang: fixed.

src/components/pacbio/PacbioLibraryEdit.vue
wendyyang: fixed.

src/services/traction/PacbioLibrary.js
wendyyang: Thank you for the suggestion. Yes, the codes is a lot cleaner by destructuring the response earlier in the chain.

src/stores/pacbioLibraries.js
wendyyang: fixed.

src/services/traction/PacbioLibrary.js
```diff
@@ -96,8 +96,12 @@ const buildLibraryResourcePayload = ({
...
   const request = useRootStore().api.traction.pacbio.libraries
   const promise = request.update(buildLibraryResourcePayload(libraryFields))
-  const { success, errors } = await handleResponse(promise)
-  return { success, errors }
+  const {
+    success,
+    body: { data: { attributes: updatedLibrary } = {} } = {},
+    errors,
+  } = await handleResponse(promise)
+  return { success, errors, updatedLibrary }
```
seenanair, line 104: Would you please update the documentaion of this method as it returns a library object instead of Promise?

seenanair (commented):
Looking good 👍

src/services/traction/PacbioLibrary.js
wendyyang: updated.

seenanair (approved):

PR #2210, opened 2025-02-20
request with status 204 code rise an JSON exception

sabrine33:


stevieing (approved):
Good spot.

Worth deploying to UAT just to check that everything is working as expected.

sabrine33:
> Good spot.
> 
> Worth deploying to UAT just to check that everything is working as expected.

I confirm the fix works as expected in UAT 

image

PR #2231, opened 2025-03-04
Y24-453  select the rebasecalling process when creating an ONT run

wendyyang:
Closes #

### add rebasecalling process attribute to ONT run model
- 

### DIFF
--- a/src/components/ont/runs/ONTRunInformation.vue
+++ b/src/components/ont/runs/ONTRunInformation.vue
@@ -37,7 +37,6 @@
 
 <script setup>
 import { computed } from 'vue'
-import { storeToRefs } from 'pinia'
 import { useOntRunsStore } from '@/stores/ontRuns'
 import useOntRootStore from '@/stores/ontRoot'
 
@@ -45,12 +44,8 @@ import useOntRootStore from '@/stores/ontRoot'
 const ontRunsStore = useOntRunsStore()
 const ontRootStore = useOntRootStore()
 
-// Extract reactive properties using storeToRefs()
-const { currentRun } = storeToRefs(ontRunsStore)
-const { instruments } = storeToRefs(ontRootStore)
-
-// Actions
-const { setInstrumentName, setState, setRebasecallingProcess } = ontRunsStore
+const { instruments } = ontRootStore
+const { setInstrumentName, setState, setRebasecallingProcess, currentRun } = ontRunsStore
 
 // Static lists
 const statesList = ['Pending', 'Completed', 'User Terminated', 'Instrument Crashed', 'Restart']
@@ -58,7 +53,7 @@ const rebasecallingList = ['5mC + 5hmC CpG-context', '5mC + 5hmC all-context', '
 
 const instrumentOptions = computed(() => [
   { value: null, text: 'Please select an instrument', disabled: true },
-  ...instruments.value.map((instrument) => ({
+  ...instruments.map((instrument) => ({
     value: instrument.name,
     text: instrument.name,
   })),
@@ -80,7 +75,7 @@ const rebasecallingOptions = computed(() => [
   })),
 ])
 
-const newRecord = computed(() => isNaN(currentRun.value.id))
+const newRecord = computed(() => isNaN(ontRunsStore.currentRun.id))
 
 // Helper function to format state values
 const formatState = (str) => str.replace(/\s+/g, '_').toLowerCase()
--- a/src/components/ont/runs/ONTRunInformation.vue
+++ b/src/components/ont/runs/ONTRunInformation.vue
@@ -37,7 +37,6 @@
 
 <script setup>
 import { computed } from 'vue'
-import { storeToRefs } from 'pinia'
 import { useOntRunsStore } from '@/stores/ontRuns'
 import useOntRootStore from '@/stores/ontRoot'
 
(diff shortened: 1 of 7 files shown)


stevieing (changes requested):

src/components/ont/runs/ONTRunInformation.vue
```diff
@@ -54,7 +64,11 @@ export default {
       const ontRunsStore = useOntRunsStore()
       return ontRunsStore.currentRun.state
     },
-
+    rebasecallingProcess() {
```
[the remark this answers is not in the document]
seenanair: @stevieing Yes, you are right. We can just use current run and access the attributes from there, I think.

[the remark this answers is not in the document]
seenanair: Also, may be this component can be moved to composition api as well?

tests/unit/components/ont/runs/ONTRunInformation.spec.js
```diff
@@ -40,6 +41,12 @@ describe('ONTRunInformation.vue', () => {
     })
   })
 
+  describe('Rebasecalling process', () => {
```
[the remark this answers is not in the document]
wendyyang: I wrote it just for consistent as the 'Instrument' and 'state' has similar tests. Not sure if we want to remove them all if not necessary?

src/components/ont/runs/ONTRunInformation.vue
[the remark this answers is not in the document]
wendyyang: refactored.

src/components/ont/runs/ONTRunInformation.vue
```diff
@@ -11,75 +11,82 @@
...
-      const ontRunsStore = useOntRunsStore()
-      return ontRunsStore.currentRun.state
-    },
+// Initialize stores
+const ontRunsStore = useOntRunsStore()
+const ontRootStore = useOntRootStore()
+
+// Extract reactive properties using storeToRefs()
+const { currentRun } = storeToRefs(ontRunsStore)
+const { instruments } = storeToRefs(ontRootStore)
```
seenanair, line 50: I dont think we need to use storeToRefs. You can use the same destructuring on line 53 for currentRun (`const { setInstrumentName, setState, setRebasecallingProcess } = ontRunsStore`) and a similar one for instruments.

wendyyang: refactored

src/components/ont/runs/ONTRunInformation.vue
```diff
@@ -11,75 +11,77 @@
...
+
+// Static lists
+const statesList = ['Pending', 'Completed', 'User Terminated', 'Instrument Crashed', 'Restart']
+const rebasecallingList = ['5mC + 5hmC CpG-context', '5mC + 5hmC all-context', '6mA all-context']
+
+// Computed properties
+const instrumentName = computed(() => ontRunsStore.currentRun.instrument_name)
+const state = computed(() => ontRunsStore.currentRun.state)
+const rebasecallingProcess = computed(() => ontRunsStore.currentRun.rebasecalling_process)
+
```
seenanair, line 58: Would it better to destructure currentRun and use it directly in v-model in in script so we can remove all these computed properties for store?  

Something like the following for `instrument_name`, `state` and `rebasecalling_process`


`const { setInstrumentName, setState, setRebasecallingProcess , currentRun} = ontRunsStore`

and in script 
```
<traction-select
          id="instrument-selection"
          :options="instrumentOptions"
          :model-value="currentRun.instrumentName"
          :disabled="!newRecord"
          @update:model-value="setInstrumentName"
        ></traction-select>

```

seenanair (commented):
Good progress, nearly there. I've shared just one thought.

src/components/ont/runs/ONTRunInformation.vue
wendyyang: if not using storeToRefs, and also remove computed currentRun properties, the test 'returns false when currentRun is not a new record' failed because the currentRun cannot automatically updated when setCurrentRun. I choose to use storeToRefs to ensure reactivity.

seenanair: Oh, ok. I see an issue with the test — it’s not replicating the exact scenario. What it’s doing is updating the store’s currentRun after mounting the component using the store method. But in reality, we always mount the component with the currentRun already set.
So, shouldn’t it work that way instead?

` it('returns false when currentRun is not a new record', () => {
      const {wrapper} = mountWithStore(ONTRunInformation, {
        initialState: {
          ontRuns: { currentRun: {
            id: 1,
            instrument_name: '',
            state: '',
            flowcell_attributes: [],
          } },
          ontRoot: { resources: { instruments: mockInstruments } },
        }
      })
      expect(wrapper.vm.newRecord).toEqual(false)
    })`

> Yes, Traction itself works fine without storeToRefs. Just the test not working which might suggest potential problem by this approach. describe('#newRecord', () => { it('returns true when currentRun is a new record', () => { expect(ontRunInfomation.newRecord).toEqual(true) }) it('returns false when currentRun is not a new record', async () => { store.setCurrentRun({ id: '1', instrument_name: 'ABC', state: '', flowcell_attributes: [], })
> 
> ```
>   expect(ontRunInfomation.newRecord).toEqual(false)
> })
> ```
> 
> }) after reset the currentRun, the currentRun is still the previous run, hence fail.

wendyyang: Yes. Your revised test works! Could you please review again? Many thanks!

seenanair (approved):
👍

stevieing (approved):

PR #2272, opened 2025-03-26
feat(ci): removes chrome version pin

BenTopping:
Testing removing the chrome version pin for CI

- Removes CI Chrome version pinned (probably to fix an old cypress bug)


seenanair (approved):
🔥 👍

Issue #2256, opened 2025-03-18
Y25-194 - [BUG] Library Batch Creation: Prevent multiple clicks and show progress feedback


seenanair:
**Describe the bug**
Following the investigation into the issue related to library batch creation, which we were unable to reproduce, it was found that the button remains enabled for a short period until the batch creation response is received. This could lead to issues such as users clicking the button multiple times.

**RT Ticket Number**
[SR-24210](https://sanger.freshservice.com/a/tickets/24210?current_tab=details)

**Expected behaviour**

- Provide feedback on the page indicating that the batch creation process is still in progress.
- Display a progress icon or show a message like "Creating library batch..." to inform users.
- Disable the button immediately upon clicking to prevent duplicate submissions.
- Re-enable the button only if the batch creation request fails.

PR #2273, opened 2025-03-26
Y25-194: add progress indicator for library creation process and prevent multiple clicks for library batch creation

seenanair:
Closes https://github.com/sanger/traction-ui/issues/2256

- Disable the 'Create Library Batch' button immediately upon clicking to prevent duplicate submissions.
- Display a progress icon or show a message like "Creating library batch..." to inform users.


src/views/pacbio/PacbioLibraryBatchCreate.vue
```diff
@@ -94,6 +94,16 @@
                 <traction-table :fields="state.csvTableFields" :items="state.csvData" />
               </div>
             </div>
+            <div class="flex flex-row px-4 space-x-2 h-2">
+              <div
+                v-if="isCreationInProgress"
+                id="progress-indicator"
```
BenTopping, line 100: if the id is just used to for tests might be better to use `data-testid`

src/views/pacbio/PacbioLibraryBatchCreate.vue
```diff
@@ -94,6 +94,16 @@
...
+                class="flex flex-row items-center"
+              >
+                <label class="text-md text-blue-400">Creating Libraries...</label>
```
BenTopping, line 103: Advisory: In other places we put the spinner in the button - e.g. https://github.com/sanger/traction-ui/blob/develop/src/components/pacbio/PacbioPoolEdit.vue#L110. Might be slightly clearer if we do the same here.

BenTopping (commented):
Nice, works well, couple minor suggestions.

src/views/pacbio/PacbioLibraryBatchCreate.vue
seenanair: Good point! I missed that. It's better to keep it consistent. 👍

BenTopping: Do you think we still need the `Creating libraries..` text too? Not sure if that can be removed or moved to the button if the spinner is enough indication.

seenanair: I thought the 'Creating libraries...' text would make it clearer to users what action is happening as though the spinner is a great visual indicator, it doesn’t explicitly tell users what is being processed.
Moving the text into the button would mean replacing the button's label dynamically, which I am not much into as the same button displaying two different button texts which serves different purposes.
Do you think the spinner is good enough, then I will remove text?

Issue #2234, opened 2025-03-05
Y25-175 - As a long read team lead (Danni) I would like to see the retention instructions for each sample


stevieing:
**User story**

As a long read team lead (Danni) I would like to see the retention instructions for each sample
**Who are the primary contacts for this story**

Steve / Danni

**Who is the nominated tester for UAT**

Danni / Long Read

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**

Consider if these features can be _featured flagged_ to decouple testing and deployment.
To be considered successful the solution must allow:
- [ ] When I visit the Pacbio sample page I should see the retention instructions for each sample
- [ ] When I visit the ONT sample page I should see the retention instructions for each sample

**Dependencies**

This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

Sequencescape only stores retention instructions at labware level. With automation it may be that long read want to see this information at plate level. We will delay this for now and review when automation comes in.

**Deployment notes**

Add any details that should be noted when deploying this feature. For example, is a database migration needed, a specifc rake task or other script that needs to be run, a dependency on another application or service that needs to be released first etc?
These notes are intended to ensure the developers releasing the feature know what needs to be done.

PR #2319, opened 2025-05-08
Y25-175 display retention instructions for sample

seenanair:
Closes https://github.com/sanger/traction-ui/issues/2234

- 

### DIFF
--- a/src/views/ont/ONTSampleIndex.vue
+++ b/src/views/ont/ONTSampleIndex.vue
@@ -40,7 +40,7 @@ import DataFetcher from '@/components/DataFetcher.vue'
 import FilterCard from '@/components/FilterCard.vue'
 import useQueryParams from '@/composables/useQueryParams.js'
 import useLocationFetcher from '@/composables/useLocationFetcher.js'
-import { locationBuilder } from '@/services/labwhere/helpers.js'
+import { formatRequests } from '@/lib/requestHelpers'
 
 const { mapActions, mapGetters } = createNamespacedHelpers('traction/ont/pools')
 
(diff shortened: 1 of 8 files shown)


src/views/ont/ONTSampleIndex.vue
```diff
@@ -73,6 +73,7 @@ export default {
         { key: 'cost_code', label: 'Cost code' },
         { key: 'external_study_id', label: 'External study ID' },
         { key: 'location', label: 'Location', sortable: true },
+        { key: 'sample_retention_instruction', label: 'Retention Instruction', sortable: true },
```
BenTopping, line 76: You could pass in a formatter key in this object, e.g. `{ key: 'sample_retention_instruction', ..., formatter: (instruction) => formatRetentionInstruction(instruction) }` instead of formatting on the displayedRequests. Two ways of doing the same thing so I think either is fine.

existing example: https://github.com/sanger/traction-ui/blob/develop/src/views/pacbio/PacbioLibraryIndex.vue#L162

seenanair: Yes, good point.
 The locationBuilder retrieves and formats the 'Location' column a bit differently by using requests and locationsData which applies on the entire array, which is why I thought it might be useful to combine both 'location' and 'retention_instruction' formatting into a shared utility thereby it can be reused across both PacBio and ONT. Hope that makes sense.

seenanair:
> Looks good.
> 
> I dont think PacBio has an e2e test but there is one for ont samples view, probably worth adding a check for retention instruction there.

create_pacbio_library_from_samples_page.cy contains test for library creation from sample in Pacbio samples page. I've updated it for checking the table values similar to ONTSample page.

BenTopping (approved):

PR #2321, opened 2025-05-09
Y24-361 - Mock reception

BenTopping:
Closes #https://github.com/sanger/traction-service/issues/1448

- Adds plate and tube mock-receptions to general reception page.

#### Additional notes / thought process

- A lot of the metadata and data setup in the mock receptions is fixed. E.g. sample retention instructions, number of wells in a plate. In order to make this customisable we could create a custom barcode scanning component instead of using the existing ones, allowing us to enter data for these values. But it is a fair amount of extra work for not much benefit so I have not done that in this story.
- I chose to add this as an extra reception in the general reception page instead of as a new 'UAT action' page or similar because it keeps the logic in one place and it is intuitive.
- I would recommend pulling down and using it to test it since it is a dev-focused feature it won't require UAT.

#### Screenshots
Screenshot 2025-05-12 at 11 09 51

### DIFF
--- a/src/lib/receptions/MockReception.js
+++ b/src/lib/receptions/MockReception.js
@@ -44,7 +44,7 @@ const fetchPlatesFunction = async ({ requestOptions, barcodes }) => {
 
     for (let i = 0; i < 48; i++) {
       const position = String.fromCharCode(65 + Math.floor(i / 12)) + ((i % 12) + 1)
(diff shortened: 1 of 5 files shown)


stevieing (approved):
Works fine for me. Agreed we don't need to do anymore to make it customisable

PR #2483, opened 2025-10-01
Y25-513 - Kinnex sample report number of donors

BenTopping:
Closes #https://github.com/sanger/traction-service/issues/1672

Related to service PR: https://github.com/sanger/traction-service/pull/1711

- Adds number of donors column to kinnex sample report.


KatyTaylor (approved):
I'm guessing this new attribute will be happily ignored when sending to the MLWH? (which is what we want)

PR #2560, opened 2025-11-24
fix(pacbio-run-well): resets well data between modal edits

BenTopping:
Relates to: [SR-47637](https://sanger.freshservice.com/a/tickets/47637?current_tab=details)

Well modal state was being leaked between well shows, this was mainly affecting editing existing runs due to 'id' being leaked, so the data being sent to the service was presenting all the wells in the same well.

- Cleans well modal state between shows


stevieing (approved):
I have approved it but I suspect there is a better way to do this as it looks like an anti-pattern as Javascript functions should be idempotent.

PR #2641, opened 2026-02-27
Y25-619 -[PR] flexible pooling to include a tube rack layout

yoldas:
Closes #

- Users can select the Tube Rack layout option on the flexible pooling page. This changes the UI to have the individual pools numbered by row in a 4x6 grid.

### DIFF
--- a/tests/unit/views/FlexiblePoolCreate.spec.js
+++ b/tests/unit/views/FlexiblePoolCreate.spec.js
@@ -146,16 +146,16 @@ describe('FlexiblePoolCreate', () => {
       // Enable; there are no pool positions.
       store.multiPoolCreateStore.multiPool.multiPoolPositions = {}
       await wrapper.vm.$nextTick()
-      expect(pipelineSelect.attributes('disabled')).toBeUndefined()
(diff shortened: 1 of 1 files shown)


src/views/FlexiblePoolCreate.vue
```diff
@@ -206,4 +209,17 @@ const provider = async () => {
...
+
+/**
+ * Dynamically determines the labware type based on the selected pooling layout.
+ * If the user selects 'Tube Rack', it returns the TubeRack24 configuration from LabwareTypes.
+ * Otherwise, it defaults to the Plate96 configuration.
+ *
+ * @returns {Object} The labware type configuration object (TubeRack24 or Plate96).
+ */
+const labwareType = computed(() => {
+  return multiPoolCreateStore.multiPool.pool_method === 'TubeRack'
```
BenTopping: Nice this looks good.

Wondering if we need a check here to prevent users switching if a pool exists greater than the size of the labware being switched to. E.g. a plate has a pool in position 25, if a user then switched to tube rack that pool would be hidden. Maybe we should add an alert and prevent the switch in that case. Any thoughts?

PR #2718, opened 2026-05-10
Y26-059 - [PR] Add Application Type to PacBio Sample Sheets

### LINKED ISSUE #2646: Y26-059 - Add Application Type to PacBio Sample Sheets
**Context**
Currently, all Traction-generated PacBio run sample sheets default to the application type “Other” when uploaded to SMRTLink. This prevents PacBio from identifying specific run types (e.g., Kinnex), limiting their ability to differentiate workflows and provide accurate reporting. Introducing structured application type selection will enable correct classification of new runs.

**User story**
As a user, I want Traction-generated PacBio sample sheets to include the correct application type, So that runs are accurately classified in SMRTLink and can be properly identified by PacBio.

James W, Ben T

James W

**Acceptance criteria**
- [ ] New well default 'Application' under pacbio well defaults for v25_1 revio smrtlink.
- [ ] Application contains the following dropdown values
 - Human WGS
 - Other WGS
 - Kinnex Single-Cell
 - Kinnex Full Length
 - Microbial Assembly
 - 3kb Amplicons
 - “Other” (fallback option)
- [ ] Validation must confirm that sample sheets upload successfully to SMRTLink with the selected application type correctly recognised.

**Dependencies**

Partial dependency for sample sheet testing on https://github.com/sanger/traction-service/issues/1813

yoldas:
Closes #

- Add application type to run well defaults for v25_1_revio


src/config/PacbioRunWellSmrtLinkOptions.js
```diff
@@ -981,6 +981,26 @@ const PacbioRunWellSmrtLinkOptions = {
         dataAttribute: 'full-resolution-base-qual',
       },
     },
+    {
+      name: 'application_type',
+      component: 'traction-select',
+      value: 'application_type',
+      label: 'Application',
+      required: false,
```
BenTopping: Note: It wont make a difference to the users but `required: true` Might be okay here because it has a default/fallback option and its not unselectable so it behaves a bit like required anyway.

src/config/PacbioRunWellSmrtLinkOptions.js
```diff
@@ -1012,6 +1032,7 @@ const defaultSmrtLinkAttributes = (run = {}) => {
     // Compute adaptive loading default based on run setting
     use_adaptive_loading: run.adaptive_loading ? 'True' : 'False',
     full_resolution_base_qual: 'False',
+    application_type: 'Other',
```
BenTopping, line 1036: Question: I assume each well can have a different application_type and that this isn't a run-level setting? Maybe one to ask the users, otherwise it might be better to do something like we do with use_adaptive_loading, where its stored on well but we keep it consistent across the run.

BenTopping (dismissed):
Looks good.

src/config/PacbioRunWellSmrtLinkOptions.js
yoldas: I would think it is a run-level setting that applies to all wells, but the option is listed under **SMRT Cell Settings** section per-well in page 28-29 of the SMRT Link User guide https://www.pacb.com/wp-content/uploads/SMRT-Link-v25.3-user-guide.pdf . I will ask if they want to specify it per-well. Thanks.

src/config/PacbioRunWellSmrtLinkOptions.js
yoldas: It should be required: true with 'Other' as the fallback although the default/fallback applies. I will change that.
You are right, SMRT Link user guide says "If blank or contains invalid values, default is Other." As a side note, this was "Custom" in v13, without a fallback behaviour note, just in case.

src/config/PacbioRunWellSmrtLinkOptions.js
yoldas: May 26, 2026. Well default and editable per well. https://psd-team.slack.com/archives/C04QUMC5DK6/p1779785780743729?thread_ts=1778494519.712269&cid=C04QUMC5DK6

tests/unit/components/pacbio/PacbioRunWellDefaultEdit.spec.js
```diff
@@ -203,4 +209,40 @@ describe('PacbioRunWellDefaultEdit', () => {
...
+
+    it('will have a selected smrt link version of v25_1_revio', () => {
+      expect(store.smrtLinkVersion.id).toEqual(smrtLinkVersions[3].id)
+    })
+
+    describe('input', () => {
+      it('sets default application_type to "Human WGS" when first option selected', async () => {
+        const options = wrapper.find('[data-attribute=default-application-type]').findAll('option')
+        // select the first option (Human WGS)
+        await options[0].setSelected()
```
BenTopping: NIT: Prefer direct name find instead of relying on options index as that may change. e.g. options.find((o) => o.value == 'Human WGS');

tests/unit/components/pacbio/PacbioRunWellDefaultEdit.spec.js
```diff
@@ -203,4 +209,40 @@ describe('PacbioRunWellDefaultEdit', () => {
...
+        expect(store.defaultWellAttributes.application_type).toEqual('Human WGS')
+      })
+
+      it('sets default application_type to "Other" when last option selected', async () => {
+        const options = wrapper.find('[data-attribute=default-application-type]').findAll('option')
+        // select the last option (Other)
+        await options[options.length - 1].setSelected()
```
BenTopping: Same as above comment.

BenTopping (changes requested):
Looks good. Tested locally in conjunction with the service PR.
- Correct options appear ✅ 
- Default well options work ✅ 
- Ability to update existing wells ✅ 
- Sample sheet shows correct options ✅ 
- When checking existing runs the wells appear invalid and application_type shows empty instead of 'Other' ❌ 
    - Think we either need to migrate existing wells to all be set to 'Other' or have a trick in the front end to show 'other' when value is nil.

Image

tests/unit/components/pacbio/PacbioRunWellDefaultEdit.spec.js
yoldas: Thanks. I have changed it to `await options.find(option) => option.element.value == 'Human WGS').setSelected()` because the option wraps the element.

yoldas:
> * When checking existing runs the wells appear invalid and application_type shows empty instead of 'Other' ❌
>   
>   * Think we either need to migrate existing wells to all be set to 'Other' or have a trick in the front end to show 'other' when value is nil.

Thanks. I have added the trick on the service side by returning the default_value of the smrt link version option when the smrt link version is  v25_1_revio . Could you please check the overridden store access on Pacbio::Well using application_type and application_type_default methods in the service PR?

BenTopping (approved):
LGTM

PR #2741, opened 2026-06-05
tests: disable vue dev tools for testing

BenTopping:
Aim to fix flakiness caused by elements occassionally being hidden by vue dev tools.

- Adds conditional vueDevTools plugin load to prevent it being present in cypress tests.


StephenHulme (approved):
Fun times ✅ 
I got caught out by the Chrome leaked-password-manager again this week too.

PR #2767, opened 2026-07-07
Y26-117 - Collect flexible pooling errors for all pools

### LINKED ISSUE #2687: Y26-117 - As a user (Tavis M) I would like the flexible pooling errors for all pools to be displayed at once to prevent repetitive reuploads
**User story**

As a user (Tavis M) I would like the flexible pooling errors for all pools to be displayed at once to prevent repetitive re uploads.

When uploading a flexible pooling file with multiple issues across multiple pools, currently only the first pool is validated and an error is displayed. This means if other pools have issues they are not seen until the first pool is fixed and reuploaded.
It is better UX to validate all the pools during the first upload and display all the errors to allow the users to fix them at once without reuploading multiple times.

Tavis M, Ben T

Tested by dev should be fine but Tavis M if not

**Acceptance criteria**

Consider if these features can be _featured flagged_ to decouple testing and deployment.
- [ ] When uploading a flexible pooling file with errors across multiple pools, the errors are all shown on initial upload

**Additional context**

This is a small change, offending area: https://github.com/sanger/traction-ui/blob/develop/src/stores/multiPoolCreate.js#L142. We need to collect errors rather than early return.

BenTopping:
Closes #2687 

- Processes all pools in the multi pool upfront and collect errors instead of early exit on first failed pool.

### DIFF
--- a/src/stores/multiPoolCreate.js
+++ b/src/stores/multiPoolCreate.js
@@ -139,6 +139,8 @@ export const useMultiPoolCreateStore = defineStore('multiPoolCreate', {
         pacbioPoolCreateStore.$reset()
         // Array to collect errors from building each pool. If any pool fails to build, we will return these errors and not store any valid pools.
         const poolBuildErrors = []
+        // Object to hold the new multi pool positions. We will only set this to the store state if all pools are built successfully.
+        const newMultiPoolPositions = {}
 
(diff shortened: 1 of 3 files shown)


StephenHulme (approved):
Looks good - much better UX 👌

PR #2771, opened 2026-07-09
Remove stale feature flags

BenTopping:
Removes stale feature flags from the code base

- pacbio_pool_auto_calculate_flag [2e27717](https://github.com/sanger/traction-ui/pull/2771/commits/2e2771779873ac4cc1750ac66150b477adfd2d83)
- flexible_pooling_flag [171b64d](https://github.com/sanger/traction-ui/pull/2771/commits/171b64dbf20390064f49f9d921046945df56be0a)
- Y26-111-user-auth [5bbc206](https://github.com/sanger/traction-ui/pull/2771/commits/5bbc2068da02302d4900e71d49576d665218b991)

### DIFF
--- a/src/components/TractionSessionButton.vue
+++ b/src/components/TractionSessionButton.vue
@@ -51,10 +51,12 @@
 </template>
 
 <script setup>
-import { onMounted, ref, computed } from 'vue'
+import { onMounted, ref, computed, inject } from 'vue'
 import { useAuth } from '@okta/okta-vue'
 
 const auth = useAuth()
+// okta-vue provides authState under the 'okta.authState' key for Composition API.
+const authState = inject('okta.authState', ref(null))
 
 const login = async () => {
   await auth.signInWithRedirect()
--- a/tests/e2e/specs/auth_flow.cy.js
+++ b/tests/e2e/specs/auth_flow.cy.js
@@ -63,10 +63,5 @@ describe('Auth', () => {
       cy.get('[data-action="logout"]').should('not.exist')
       cy.get('[data-action="login"]').should('be.visible')
     })
-
-    it('cannot access protected pages', () => {
-      cy.visit('#/labwhere-reception')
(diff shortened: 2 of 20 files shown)


tests/e2e/specs/auth_flow.cy.js
```diff
@@ -63,10 +63,5 @@ describe('Auth', () => {
       cy.get('[data-action="logout"]').should('not.exist')
       cy.get('[data-action="login"]').should('be.visible')
     })
-
```
BenTopping, line 66: I removed this test because okta attempts to sign in with redirect when you visit a protected page when not logged in. Its a bit of a mocking nightmare and tests the okta library more than anything which we dont need to do.

src/components/TractionSessionButton.vue
```diff
@@ -51,10 +51,12 @@
 </template>
 
 <script setup>
-import { onMounted, ref, computed } from 'vue'
+import { onMounted, ref, computed, inject } from 'vue'
 import { useAuth } from '@okta/okta-vue'
 
 const auth = useAuth()
+// okta-vue provides authState under the 'okta.authState' key for Composition API.
+const authState = inject('okta.authState', ref(null))
```
StephenHulme, line 58: note: does not appear to be related to feature flags

StephenHulme (approved):
Love removing code ✂️ looks good 👍

src/components/TractionSessionButton.vue
StephenHulme: (I'm guessing this is related to the okta test changes you commented on? - Sorry, I only saw that comment after adding mine)

BenTopping: authState is used in the component but it was not actually defined anywhere. It was made available globally by the okta-vue package. This is just explicitly declaring where we are getting the value from.

StephenHulme: Ah... thanks for the details - sounds good.

